PR #8, opened 2023-03-07
Link do-while, remove custom blocks & update enum-variants-strings

kaleidawave:
Links up the `do` token parsing to the `do { ... } while (...)` structure (issue found during fuzzing):

https://github.com/kaleidawave/ezno/blob/97098d1d9e6ce4bcef98eed29e8c03712c45f4e4/parser/src/statements/mod.rs#L219-L221

This was previously left out because I wanted to handle (as an extra) the `do` expression based on the [tc39 proposal](https://github.com/tc39/proposal-do-expressions). However I don't think it offers anything better that IIFE. So have removed that code as well as some really old other custom block syntax.

Also updated to latest [enum variants strings](https://crates.io/crates/enum-variants-strings) as it had some fixes for changing modes in the ast explorer CLI


kaleidawave:
Ooof, might want to re-read the PR body. Maybe I will have a change of heart / or it actually gets added to the spec and add then I will add it back. For now it didn't make sense to keep it around without it being properly handled.

jasikpark:
lol yeah i read that

PR #15, opened 2023-03-22
Better cjs compatibility

markthree:
only need an `npm run build` to generate a cjs plug-in


kaleidawave:
Thanks for this PR. Also for fixing the README, missed that oops.

Would be good to get an explanation what this fixes/unlocks. I not particularly up to date on the CJS ESM stuff but wondering what situations you use the CJS export for a package that is written in ESM?

markthree:
Another reason is that there may be package authors who refer to this library but want to package cjs.

Finally, you may only be able to package inline dependencies or use `import('unplugin-ezno')` to load asynchronously。

markthree:
> Awesome, thanks for the clarification. So this basically allows the plugin to be used in CJS environments, which is a really useful change. Some things:
> 
> There are two JS based libraries for Ezno. The "main library and cli" (`js-cli-and-library`) and the "JS based plugin" `js-based-plugin` that supports esbuild, webpack are more thanks to unplugin. So...
> 
> * Could this change be added for the main JS library (https://github.com/kaleidawave/ezno/tree/main/src/js-cli-and-library). It is the main entry point for using Ezno as a dependency of another JS package. Also with this change you should check that `cli.js` still works
> * This PR is currently for the `unplugin` based plugin which imports the ezno js library. Is the still a benefit for it being CJS compatible if it is only for directly importing in an unplugin supported build tool?
> 
> Also another small thing before a merge. Would it be possible to keep the flat style of the files. Just a small thing to maintain to keep consistency with the rest of the project and to avoid clicking through a `src` folder to find a single small file. Just took a look at `unbuild` and looks like you may be maybe able to do it though a config section in the `package.json`!

Fine, I'll improve it tomorrow 🥳

markthree:
All packages are already compatible with cjs 😁

kaleidawave:
Awesome, thanks for the changes. Configuring it in package.json is neat. Will check it out later and merge.

_assume you mean **now** CJS compatible rather than already 😄_

kaleidawave:
Hey, checked it out and found a few issues. `bindings.js` doesn't seem to be included and a few things.

Not to worry though, I can figure them out from here. I will change the current `wasm-bindgen` from `--target nodejs` to `--target bundler` as that is the process now. Then need to use a rollup plugin that handles the WASM... etc. I will try and fit Deno and web support into this PR at the same time

markthree:
> Hey, checked it out and found a few issues. `bindings.js` doesn't seem to be included and a few things.
> 
> Not to worry though, I can figure them out from here. I will change the current `wasm-bindgen` from `--target nodejs` to `--target bundler` as that is the process now. Then need to use a rollup plugin that handles the WASM... etc. I will try and fit Deno and web support into this PR at the same time

Uh huh, I don't know much about Rust's wasm 😬

kaleidawave:
Hey @markthree. Thanks for the PR, I had never heard of unbuild so it's been interesting to learn that tool. Your additions were a great base to build on. I made some changes to the Rust code so that it takes in FS and CLI access as parameters. This made it a lot more flexible and so made it easier to add these improvements to the JS side. It should be building the CJS and MJS modules correctly now.

You should be able to do this now.
```js
const {build} = require("ezno/initialized");
```
I just realised there are a few more changes (I left the readme out of date 🤦‍♂️), so will make them in the morning before I merge. Let me know if it is all good and whether I missed anything.

kaleidawave:
CI Green, PR Merged!

markthree:
> CI Green, PR Merged!

nice，awesome 
SoExcited~GIF

PR #16, opened 2023-03-26
Improve CI testing and publishing

### LINKED ISSUE #7: Improve CI testing and publishing
### Testing
- **Run fuzzing tests in parallel**. Will show red light correctly and be faster
- Get cargo fuzz from: https://github.com/rust-fuzz/cargo-fuzz/releases/tag/0.11.2
- Split up testing step by crates (eventually this should mean that it should only run tests for crates that have changed)
- Have some sort of feature matrix for testing
- **Add hyperfine and flamegraph** performance tracing
- Enforce Cargo.toml formats

*need to check whether caching is working correctly*

### Publishing
- Automatically add `ezno-ast-generator` if `ezno-parser` is updated

kaleidawave:
Close #7


jasikpark:
https://github.com/kaleidawave/ezno/actions/runs/4526106473/jobs/7971052826?pr=16 seems like an invalid config?

.github/workflows/rust.yml
```diff
@@ -16,51 +32,98 @@ jobs:
...
+      - name: Run base tests
+        run: cargo test
 
   fuzzing:
     needs: validity
     runs-on: ubuntu-latest
+    continue-on-error: true
+    strategy:
+      matrix:
+        fuzz-type: [module_roundtrip_naive, windows-latest]
```
jasikpark, line 88:
```suggestion
        fuzz-type: [module_roundtrip_naive, module_roundtrip_structured]
```

I would also maybe rename that to `fuzz-target` since that's more accurate once more functions are fuzzed

kaleidawave: Definitely wasn't copying from the GH actions matrix documentation 🙈. `fuzz-target` is a much better name, I wasn't too confident with `fuzz-type`. Will change

kaleidawave:
The caching has improved the times a lot 😁
image

PR #23, opened 2023-06-08
Add parsing for Try Catch statements

### LINKED ISSUE #10: Try catch AST
Somehow along the way have missed out parsing try catch (finally) statements: https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/try...catch

https://github.com/kaleidawave/ezno/blob/d311155bcb4277843551bbf6b7421454c023c149/parser/src/statements/mod.rs#L146-L148

This should be very similar logic to statements like `if` and `while`, **however the spec says that statements have to be `Block`s and not allow single statements**.

The [while statement code](https://github.com/kaleidawave/ezno/blob/main/parser/src/statements/while_statement.rs) is a good reference for how it should be implemented. Can copy that and then adjust for the keywords and such. Then just need to import that back into `statements/mod.rs`, create a similar variant and hook it up to it's parse, position and `to_string` implementation.

If anyone wants to tackle this and needs more information/help, let me know 👍

nathanbabcock:
Closes #10 

This project is really awesome. Hoping to get more familiar with it. So far the structure has been very intuitive and easy to follow. Thanks!

-NB

### DIFF
--- a/parser/src/statements/try_catch_statement.rs
+++ b/parser/src/statements/try_catch_statement.rs
@@ -1,1 +1,59 @@
+use std::borrow::Cow;
 
+use crate::{ASTNode, Block, TSXKeyword, TSXToken, VariableIdentifier};
+use source_map::Span;
+use visitable_derive::Visitable;
+
+#[derive(Debug, PartialEq, Eq, Clone, Visitable)]
+#[cfg_attr(feature = "self-rust-tokenize", derive(self_rust_tokenize::SelfRustTokenize))]
+pub struct TryCatchStatement {
+	pub inner: Block, // both try and catch blocks must always be braced
+	pub error_var: VariableIdentifier,
+	pub catch_inner: Block,
+	pub position: Span,
+}
+
+impl ASTNode for TryCatchStatement {
+	fn get_position(&self) -> Cow<Span> {
+		Cow::Borrowed(&self.position)
+	}
+
+	fn from_reader(
+		reader: &mut impl tokenizer_lib::TokenReader<TSXToken, Span>,
+		state: &mut crate::ParsingState,
+		settings: &crate::ParseSettings,
+	) -> Result<Self, crate::ParseError> {
+		let start_span = reader.expect_next(TSXToken::Keyword(TSXKeyword::Try))?;
+		let inner = Block::from_reader(reader, state, settings)?;
+		reader.expect_next(TSXToken::Keyword(TSXKeyword::Catch))?;
+		reader.expect_next(TSXToken::OpenParentheses)?;
+		let error_var = VariableIdentifier::from_reader(reader, state, settings)?;
+		reader.expect_next(TSXToken::CloseParentheses)?;
+		let catch_inner = Block::from_reader(reader, state, settings)?;
+		Ok(Self {
+			position: start_span.union(&catch_inner.get_position()),
+			inner,
+			error_var,
+			catch_inner,
+		})
+	}
+
+	fn to_string_from_buffer<T: source_map::ToString>(
+		&self,
+		buf: &mut T,
+		settings: &crate::ToStringSettings,
+		depth: u8,
+	) {
+		buf.push_str("try");
+		settings.add_gap(buf);
+		self.inner.to_string_from_buffer(buf, settings, depth + 1);
+		settings.add_gap(buf);
+		buf.push_str("catch");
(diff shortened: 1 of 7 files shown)


parser/src/statements/try_catch_statement.rs
```diff
@@ -1 +1,59 @@
...
 
+use crate::{ASTNode, Block, TSXKeyword, TSXToken, VariableIdentifier};
+use source_map::Span;
+use visitable_derive::Visitable;
+
+#[derive(Debug, PartialEq, Eq, Clone, Visitable)]
+#[cfg_attr(feature = "self-rust-tokenize", derive(self_rust_tokenize::SelfRustTokenize))]
+pub struct TryCatchStatement {
+	pub inner: Block, // both try and catch blocks must always be braced
+	pub error_var: VariableIdentifier,
```
kaleidawave, line 11: Just checked the spec and there are a few things additional things around `error_val`
- It can be missing entirely (parenthesis and all)
- It can include destructuring `catch ({a, b})`, so can use [VariableField](https://docs.rs/ezno-parser/0.0.3/ezno_parser/enum.VariableField.html) which adds the additional destructuring variants
- [TypeScript allows an optional annotation here](https://stackoverflow.com/a/69577476/10048799), so can use [TypeReference](https://docs.rs/ezno-parser/0.0.3/ezno_parser/types/type_references/enum.TypeReference.html)
So `error_val: Option<(VariableField, Option<TypeReference>)>` would be great

parser/src/statements/try_catch_statement.rs
```diff
@@ -1 +1,59 @@
...
+	pub catch_inner: Block,
```
kaleidawave, line 12: [The spec says catch clause can be missing and there is additionally the finally block](https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/try...catch#description). (I couldn't tell you what the `finally` block is used for 😂, need to look when you would use that)
So it should be
```
catch_inner: Option<Block>,
finally_inner: Option<Block>,
```

kaleidawave (commented):
Awesome, these are really useful changes! The code is perfect and adding the test is also great. There are just two comments for some minor changes to support the full try...catch statement. 

`if let Some(Token(TSXToken::Keyword(TSXKeyword::Catch), _)) = reader.peek()` will come in useful for testing whether the next token is some value before 'eating' it / advancing the lexer. Don't worry about a case happening where both blocks are both `None`, I will fix that at some point!

nathanbabcock:
Partially updated, will revisit tomorrow...

I'm torn on whether to use an enum (making illegal states unrepresentable, like a `try` without any `catch`/`finally`, or an `exception_var` with no `catch`), or just keep things simple with a flat struct allowing any possible combination of `Option`'s.

Any preference?

kaleidawave:
Hmm yeah representing the different cases is difficult. 

Just thinking and I have a [enum I was playing with in the checker](https://github.com/kaleidawave/ezno/blob/422a3591c9b577efd00843fba9036b3a90a678b7/checker/src/utils.rs#L124) which is 'inclusive either'. Effectively `(Option<T>, Option<U>)` **without** the `(None, None)` case. **So I think it is best to go with a simple flat struct**, then at some point maybe I make the 'inclusive either' util type in to a crate and can swap it out so `try { ... }` on its own is unconstructable.

Aside from that, the new error is looking good and the test for all the variants is awesome!

nathanbabcock:
Cool, I dig it.

One other thing to mention: I added the [`pretty_assertions`](https://docs.rs/pretty_assertions/latest/pretty_assertions/) crate as a dev dependency to help me quickly visualize the string diffs coming out of the try-catch tests. If you already have another preferred method instead, we can revoke that [commit](https://github.com/kaleidawave/ezno/pull/23/commits/23f647a0f67698fc30f4d42fbd4746f084a9773b).

kaleidawave:
Fuzzing is broken but not to worry. `naive` shows an existing error and `structured` looks like a issue with the actual fuzzer. Will fix in a different PR.

Everything else looks great. Well done with `WithComment`, I forgot that could be a case. Normally a bit sceptical of adding dependencies, as trying to keep them to minimum in Ezno. But this looks like a super useful crate and as it is only for testing that's fine!

nathanbabcock:
Woohoo! Thanks. 😀

Issue #43, opened 2023-07-24
Lexing of spaces breaks attribute parsing


techieAgnostic:
When evaluating an expression that has an attribute, the current lexer in the parser-fixes2 branch allows matching spaces as part of the tag name. This prevents the Lexing State from moving to lexing the attributes, instead giving an error when it reaches the `=` character.

Example: running `ezno ast-explorer ast` and using the expression `<div className="test"></div>`
Output:
```
error: 
  ┌─ INPUT:1:2
  │
1 │ <div className="test"></div>
  │  ^ Invalid character '=' in JSX tag
```

When removing the match on spaces [here](https://github.com/kaleidawave/ezno/blob/parser-fixes2/parser/src/lexer.rs#L543) it allows the lexer to continue onto the attribute lexing state, giving the following output for the same example:
```
JSXRoot(
    Element(
        JSXElement {
            tag_name: "div",
            attributes: [
                Static(
                    "className",
                    "test",
                    5..21,
                ),
            ],
            children: Children(
                [],
            ),
            expression_id: ExpressionId(1),
            position: 0..28,
        },
    ),
)
```

I have a fork with the changes made to the parser-fixes2 branch [here](https://github.com/techieAgnostic/ezno/tree/parser-fixes2), but wanted to double check that there wasn't a reason to be lexing spaces as part of the tag name first.

Thank you :)

PR #44, opened 2023-07-25
Update lexer.rs

techieAgnostic:
Fix for Issue #43


kaleidawave:
Thanks for finding the issue and figuring out what caused it. It is definitely an error, I must have messed up copying and pasting logic or smth.

I forgot what I was doing with the `parser-fixes2` branch but it is a bit outdated. Do you think you could rebase (or whatever the term is) this PR so it is targeting the `main` branch?

PR #49, opened 2023-08-24
More parser fixes

kaleidawave:
private fields, static blocks and others


kaleidawave:
Seems like boa adds semicolons occasionally, so fuzzing still failing. Else looks okay

PR #61, opened 2023-09-15
Tokens use start positions rather than Spans

kaleidawave:
- AST no longer has source id (new change incoming to support bundling)
- Use GetFieldByType derive macro occassionally
- Every ASTNode has a fixed position that is NO LONGER CALCULATED


kaleidawave:
Additionally added
- Some fixes in the parser (`>>>=` wasn't handled, `MultipleExpression` in `for in` and switch expressions, import & export statements)
- All AST nodes can be serialized to JSON
- `parse_expression` in the JS library
- `lex` command which prints coloured tokens (although has some problems)
- `SourceId` on environment

Issue #63, opened 2023-09-15
Read from imports


kaleidawave:
Imports are not currently implemented :/

They should read from the **JavaScript** source 99% of the time (Ezno has some problems with existing `.d.ts` files, JS is a better source of truth). 

To do
- Think about how import resolving works (look at whether there is an existing library that can do this for us)
- Have a `Module` struct that contains information (I think there is one from my previous attempt, which needs improving)
- Think about the side-effects of module important
- Record things around `export` declarations
- Dynamic import?

kaleidawave:
Marking this as 90% completed by #71. Separate issues are:
- Name lookup, needs `package.json` and `node_modules` reading. Also want to support Deno packages here (not sure how they work tho)
- Dynamic / expression level `import`. Will have to rule out some cases when the import is too dynamic

chengyuhui:
Is it possible to further abstract this (for example, with a `ModuleResolver` trait) so that it is possible to use projects like `oxc-resolver` to handle `package.json` and `node_modules`?

```rust
trait ModuleResolver {
  // `path` is the full path to the file that is trying to import `spec`.
  fn resolve<P: AsRef<Path>>(&self, path: P, spec: &str) -> Result<(Path, String)>;
}
```

kaleidawave:
Yep current there is the [`ReadFromFS`](https://docs.rs/ezno-checker/latest/ezno_checker/trait.ReadFromFS.html) trait that is `Path -> String`. At the moment only relative path imports are supported. [The resolving for that (aka joining relative parts) is done internally in the checker](https://github.com/kaleidawave/ezno/blob/182dc1f031d26860c52c70b25cdf02bfbdc5941f/checker/src/lib.rs#L316-L319). 

When I get to adding external package support (which needs more things to work first) I think I will add `oxc-resolver` internally and extend that logic using `oxc-resolver`s resolving algorithm.

Maybe I that should be independent of `ezno-checker` and so it should be done in the CLI instead (for example is `package.json` parsing part of the JavaScript specification, is it different for Deno etc)? Not sure yet...

--- 
For commonjs, I think `require` could be implemented as a [*constant function*](https://github.com/kaleidawave/ezno/blob/main/checker/src/behavior/constant_functions.rs), which is a custom piece of Rust code that can do internal things that aren't possible (or very difficult) to represent in the type system. So as long as it isn't a dynamic path (which breaks everything) it should eventually be possible

chengyuhui:
[...]

PR #71, opened 2023-10-30
File importing and exports

kaleidawave:
Basic support for cross file checking. Implements #63

- Adds the ability to break current checking and import a file during hoisting. 
- Automatically handles existing checked modules
- Handles extensions
- Abstracts existing `Variable` with additional `Import` variant


kaleidawave:
Left to do:
- [ ] Import and export types
- [ ] Side effects?
- [ ] Import default custom object
- [ ] Export from
- [ ] Tests for not found / invalid syntax / invalid import / reading export

PR #88, opened 2023-11-21
[Refactor] Removing most of clippy warnings

JulesGuesnon:
# Description

This PR applies most of the suggestions made by clippy.
There's still some warnings that I didn't take care of:
- [`Result<_, ()>`](https://rust-lang.github.io/rust-clippy/master/index.html#/result_unit_err) -> I think it may needs to define how to do the error reporting, or if this warning can be disabled
- [Function with too much arguments](https://rust-lang.github.io/rust-clippy/master/index.html#/too_many_arguments) -> Again, either "a big" refactor is needed, or to disable the warnings

Also, I think this PR could be a good timing to choose a lint level and explicitly add it everywhere

### DIFF
--- a/checker/examples/check.rs
+++ b/checker/examples/check.rs
@@ -14,7 +14,7 @@ fn main() {
 		path.to_path_buf(),
 		HashSet::from_iter(std::iter::once(ezno_checker::INTERNAL_DEFINITION_FILE_PATH.into())),
 		|path: &std::path::Path| {
-			if path == &PathBuf::from(ezno_checker::INTERNAL_DEFINITION_FILE_PATH) {
+			if path == PathBuf::from(ezno_checker::INTERNAL_DEFINITION_FILE_PATH) {
 				Some(ezno_checker::INTERNAL_DEFINITION_FILE.to_owned())
 			} else {
 				fs::read_to_string(path).ok()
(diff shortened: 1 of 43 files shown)


kaleidawave:
Cool. Changes look good so far!

With the `Result<..., ()>` those are functions where I haven't quite finished what the error should be. Cool to know they are identified by clippy.

In terms of the too many function parameters. I was considering bundling `ThisValue`, arguments and other argument kind of things into a struct. That would squash a few parameters into one to fix the lint. If you want to attempt then that would be great 👍 

Also slightly more ambitious but I am also thinking about putting `Publicity` into ProperyKey::String. (You can't have non statically known private keys so should be fine to not be on PropertyKey::Type). That should solve the property key too many parameters. 

Don't have to and could merge RN but you definitely seem capable :)

In terms of lint levels, do you have any suggestions. I think there is a recently added way to specify them in Cargo.toml. Did you encounter a common rule that you think or should be changed from default? Any other changes to the current default config? I haven't too much experience with clippy to know :)

kaleidawave:
Going to merge this now as I think it is important is added asap. The other points can be addressed in the future with other PRs

JulesGuesnon:
Hey @kaleidawave ! Sorry for the delay, I've been quite busy last days so I couldn't have a look to your feedback. I'll be busy until the end of the week I think, but I'll open another PR to address everything you mentioned when I can

kaleidawave:
No worries, I thought so 👍. Wanted to work on the project today so was keen to have the improvements in today to not end up with a bunch of merge conflict.

Issue #74, opened 2023-10-31
Expected type during value synthesis


kaleidawave:
Currently the `mapper` functions `a` parameter is synthesised as generic
```ts
declare const array1: Array<number>;
array1.map(function mapper(a) { return a + 2 })
```
then the function is checked based on the restriction.

A better way would be to synthesise the function with the knowledge/expectation that it is expected to be a function as `number => U`. This means that the `new_function` can then create parameter restrictions eagerly.

This could be considered as traditional type inference

kaleidawave:
Added the parameter-argument part with

https://github.com/kaleidawave/ezno/blob/1aa77e2aabb9533fb4fe90fa1ac9aaa24136256b/checker/src/types/calling.rs#L53-L98

Not great position + overhead of clone. But works ATM

These work though

![image](https://github.com/kaleidawave/ezno/assets/26967284/fef7f5ec-01a0-4ee6-921d-6fc22fdc855a)

[from](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#expected-parameter-from-variable-declaration)

Issue #60, opened 2023-09-13
Record possibly mutated variables / objects and do checking in effects


kaleidawave:
Given something like

```ts
function (cb: Function) {
    const obj = { a: 2 }
    cb(obj)
    console.log(obj.a)
}
```

This might fail if the function passed as `cb` deletes `a` from `obj`.

As there is no annotation syntax for marking a function as pure/without side effects (at the moment 👀) the safety isn't known inside the function. 

*This currently isn't caught in Ezno, so needs fixing*.

Instead:
- Objects passed to functions (and some other items that might be mutated by calling a unknown function) must be marked as *unstable* or something
- Using this *unstable poly types* should require checking in `apply_effect` (as it evaluates side effects and is aware of what stage the function is at)

kaleidawave:
#95 [starts on this](https://github.com/kaleidawave/ezno/blob/0881454c9de4579273d4941f8af29c33f99af34c/checker/specification/specification.md#mutating-an-object-by-a-function) might need to identify the edge cases but the basic principal is there

PR #95, opened 2023-12-05
Add checking for basic loops (while, for with condition, ...)

### LINKED ISSUE #73: For / while loops
- [ ] Add iterators for array based on length etc
- [ ] Add effects running with some iteration limit

kaleidawave:
Fixes #73

To implement
- [x] while
- [x] do while
- [x] For loops (conditional)
- [x] `break` and `continue`
  - [x] labels
  - [x] context checking
- [x] Non unrolled edition
- ~~For loops (`of` and `in`)~~


kaleidawave:
Realised that a the *unknown updates* after a loop that dependent on a poly type is the same problem as #60 . So need to do some more thinking about how to tackle that

kaleidawave:
For of will require symbols as well...

kaleidawave:
👑 
image

(also made some progress on #60)

Still some more to do though

kaleidawave:
Should be just `for of` and `for in` left to do, as well as some more tests

kaleidawave:
Think this is okay to merge. Up to 129 specification tests passing now !! 

**IMPORTANT**: This PR does not implement `for in` and `for of`. There is some technicalities (symbols, non constant objects etc) to resolve that would cause this PR to go on forever. So going to implement them after taking a break from iteration code.

Also although Array methods `.map`, `.filter`, `.includes` do work, they are a bit flaky and I don't think they should be default while both the binary types and event cache is still out and #74 doesn't work for function arguments. So leaving them out for now.

Added more constant functions, which should be useful for debugging.

PR #105, opened 2024-01-05
Some fixes

kaleidawave:
- Hopefully fixes some GH workflows
- Adds option for expression mappings in the checker
- Adds the specification to the workspace (again not sure if that is correct thing to do)

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -1,7 +1,12 @@
 name: GitHub release
 
-on: 
-  workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      ezno-version:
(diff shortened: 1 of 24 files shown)


kaleidawave:
Update: Added a catch for recursion (which would cause the stack overflow the checker) a while back

https://github.com/kaleidawave/ezno/blob/1aa77e2aabb9533fb4fe90fa1ac9aaa24136256b/checker/src/types/calling.rs#L502-L515

Again more work to do. Maybe the learnings of the [unknowns iteration counts](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#while-loop-with-unknown-number-of-iterations) could help here?

PR #110, opened 2024-01-29
Add cache back in, for-in and many other improvements

kaleidawave:
- Added `for (... in ...)`
- Added facts merging
- Created facts chain trait and use in it places for getting properties
- Added caching (binary type definitions)
  - Only allows once cache at the moment
  - Only works for one definition file (start of the world, cannot interweave files)
  - Now on by default for the internal definition
  - Reduces time to startup
- Created generic chain
- Improved object constraints
- Add `Contributions`
- Added lookup generics (for untyped collection types)
- Started on mapped types
- Fix for type aliases and object types
- *Interrupting events* are stored on `Environment`s and that is now where return types and thrown values are looked
- Added `Array.map`, which works sometimes!
- *Probably some others which I forgot while I was on the 2 week break*

Now passes [180](https://www.youtube.com/watch?v=fKMwsRUwWzI) specification tests 🎉


kaleidawave:
Currently doing some research in contravariance and covariance with respect to explicit generics. Not impossible but generics and structural type systems don't mix nice. Especially as my implementation tries to collect generics at the same time as subtyping of arguments against parameters (rather than two passes). We will get there!

kaleidawave:
https://www.youtube.com/watch?v=fKMwsRUwWzI

image

kaleidawave:
Just seen the date and diff count of this PR. Still lots to do but to prevent this dragging on forever (it was originally just to add the cache back), will merge now and open issues for the *leftovers*.

PR #111, opened 2024-01-30
Improve parser

kaleidawave:
##### Done

- [x] Fix `[,,]` output dropping an empty
- [x] Import assertions
- [x] Remove variable field optional expression confusion
- [x] Test declaration files (including our own for the WASM build), see [issues here](https://gist.github.com/H-Plus-Time/fef7b91f2742fc0aed3dc6722756bb22?permalink_comment_id=4885532#gistcomment-4885532)
- [x] Fix weird single statement function with call (`do function x(){}(); while (true);`)
- [x] Fix the new issue from the specification
- [x] Make all tests run on `main` branch...?
- [x] `clippy`
- [x] Fix cfg related code issue
- [x] Allows variable field for spread expression under extras
- [x] Allow `public` for constructor things under `full-typescript`

##### Not great stuff introduced
- Parse error for large exponentials on numbers (rather than clamping the size of the number)


kaleidawave:
Should look at the issue found here: https://gist.github.com/H-Plus-Time/fef7b91f2742fc0aed3dc6722756bb22?permalink_comment_id=4885532#gistcomment-4885532

kaleidawave:
💚💚💚
image

kaleidawave:
Just left is some `TypeDeclarationModule` stuff (might remove and have it all under `Module`). The following should parse **only** in type definition modules (it is the last thing to fix from the definition file generated in #114)
```ts
export default function x(): string;
```
Need a special `default` export variant which is an expression function with an optional body (normally all expression functions have to have a body but not in this case)

kaleidawave:
Lesson learnt: 10 minutes of sequential fuzzing runs doesn't run the same amount as 2 minutes of parallel fuzzing

kaleidawave:
The fuzzer has something personal against me

image

PR #126, opened 2024-03-15
More checker features

kaleidawave:
- Class hoisting and now using as internal definition (in `full.d.ts`). 
- Package imports (looking in `node_modules`). Uses `simple_json_parser`
- Change to `FunctionType` for effects

##### TODO

- [ ] class tests are not passing :(
- [ ] more hoisting
- [ ] finishing mapped types
- [ ] fix internal array methods that broke :(


kaleidawave:
Got to the bottom of why `.github/workflows/performance-and-size.yml` wasn't working. Turns out `scc .\demo.ts` wasn't working on Linux (Windows filepaths are borkt). Fixed now!

Issue #112, opened 2024-01-30
Check using the type annotation for default parameter value and try-catch variable


kaleidawave:
*putting these together because they are pretty much the same implementation*
#### Default parameter value
currently no exception is raised here
```ts
function x(a: number = "hi") {}
```
it should raise an exception for the default parameter value, as the RHS is not subtype-able under `number`

the code is here, using `type_is_subtype` with `parameter_constraint` against `value` (default_value) and sending the result to `checking_data.diagnostics_container` with the position using `A::expression_position`

https://github.com/kaleidawave/ezno/blob/b6ead3efaebbfa8c1692f296e30860b3be10f17c/checker/src/features/functions.rs#L136-L147

#### Catch variable
currently no exception is raised here
```ts
try {
	throw 5
} catch (e: string) {
	
}
```
it should raise an exception here `5` is RHS is not subtype-able under `number`

the code should go here https://github.com/kaleidawave/ezno/blob/b6ead3efaebbfa8c1692f296e30860b3be10f17c/checker/src/synthesis/statements.rs#L250

#### The shape of code
Here you can see how it is done for `x satisfies Type`. You simply want to import the `type_is_subtype` function and give the type from the annotation as the LHS type/first argument, then follow the rest.

https://github.com/kaleidawave/ezno/blob/b6ead3efaebbfa8c1692f296e30860b3be10f17c/checker/src/types/subtyping.rs#L760-L764

and emitting the error diagnostic

https://github.com/kaleidawave/ezno/blob/b6ead3efaebbfa8c1692f296e30860b3be10f17c/checker/src/lib.rs#L382-L400

---

**You want to create custom `TypeCheckErrors` for both these scenarios, to annotate that it was a `InvalidDefaultValue` or `ThrowDoesNotMatchCatchVariableAnnotation`**

You also want to add tests to `specification.md` for the cases above.

As with any issue, if you want more help or questions feel free to ask

n-keerthi-gayathri:
@kaleidawave please assign me this issue

CharlesTaylor7:
@kaleidawave @n-keerthi-gayathri 
Hey, I'm interested in taking on this task as my first contribution. I'll be following the detailed steps in the issue description, but I'll post again if I run into any issues.

PR #131, opened 2024-04-13
Check catch argument type annotation

CharlesTaylor7:
Part 1 of implementing #112

Just handles invalid catch annotations. I'll do a separate PR for checking default function arguments.

### DIFF
--- a/checker/specification/specification.md
+++ b/checker/specification/specification.md
@@ -1668,7 +1668,7 @@ try {
 
 - Expected string, found 3
 
-#### Catch annotations
+#### Catch annotation
 
 ```ts
 try {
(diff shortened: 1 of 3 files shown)


kaleidawave:
Thanks for taking this on! Looking forward to default parameter value check as well!

I think you might have missed using the new diagnostic though 😅

https://github.com/kaleidawave/ezno/pull/131/files#diff-0d883820650b8c390ce4cec20ce7c37e02118eb23085c447d0452c29f1b8229cR328

Also could the error message be `Cannot catch type {found} because the try block throws {expected}`?

Otherwise perfect 😍
```
test result: ok. 201 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.11s
```

CharlesTaylor7:
@kaleidawave Done and done.

This is an awesome project, by the way, really hope it takes off. Thanks for the quick feedback!

PR #132, opened 2024-04-13
Check default parameter value type

CharlesTaylor7:
Part of #112 
Default parameter type checking.

### DIFF
--- a/checker/specification/specification.md
+++ b/checker/specification/specification.md
@@ -315,6 +315,17 @@ function func(a: number) {
 
 - Expected string, found number
 
+#### Default parameter value
+
+```ts
+function outer(a: number) {
+    function inner(b: string = Math.floor(a)) {
+    }
+}
+```
+
+- Cannot use a default value of type number for parameter of type string
+
 #### (simple) return type checking
 
 ```ts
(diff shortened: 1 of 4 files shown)


kaleidawave:
Another great PR. Specification test is great and diagnostic great 👍. Up to 202!

Just added two quick commits on top 
- Firstly the merge conflict in `diagnostics.rs` from the first PR. 
- Secondly  I swapped the `ASTNode::get_position` call for `A::expression_position` for getting the position of where to emit the type checking error. This is because the type checking API is meant to be agnostic from a parser implementation (so in the future any other project can reuse the checking APIs without them to being specific to the `ezno-parser` parser I built). `A::expression_position` comes from `ASTImplementation` which works in under any AST definitions that implement it. For the `ezno-parser` AST definition it does use [`ASTNode::get_position`](https://github.com/kaleidawave/ezno/blob/24e35b98da5a6c705be3cce7eb56dc17bfcb32a4/checker/src/synthesis/mod.rs#L111-L113). That [fixes why the CI error on the extras tests](https://github.com/kaleidawave/ezno/actions/runs/8679227457/job/23797534034#step:11:48) as the only place you can refer to `ezno-parser` in the checker is under the synthesis directory (which mostly connects APIs to the AST definitions). Little bit complex, need to document the `ASTImplementation` trait and that whole system better!

Issue #134, opened 2024-04-14
Bug: Unsound checking of function calls with array parameters


CharlesTaylor7:
There's some bugginess with functions that declare array parameters. 

Bugs are listed with a comment ❌ .

```ts
function callWithNum(a: number) {}
function callWithArray(a: number[]) {}
function callWithArrays(...args: number[][]) {}

callWithNum(1) // ✅ type checks
callWithNum([1]) // ✅ does not type check

callWithArray(1) // ❌ type checks but shouldn't
callWithArray([1]) // ✅ type checks

callWithArrays(1, 2, 3) // ❌ type checks but shouldn't
callWithArrays([1], [2], [3]) // ✅ type checks
```
---

I'm not sure what's going on here. I initially thought the bug had to do with just the implementation of rest parameter constraint checking, but realized the issue applies to regular parameters too.

kaleidawave:
Ooof looks like a issue in subtyping logic. The `type_is_subtype` function seems to be returning that `3` has all the *properties* to be passed as a `Array<number>`, which isn't true (`3` does not have a `length` property for example so isn't the same as `Array`). The issue is in `type_is_subtype` as it also affects other places where subtyping is checked.

![image](https://github.com/kaleidawave/ezno/assets/26967284/a78576e0-146c-4eee-86a3-d7640e44d314)

kaleidawave:
Running with `$Env:EZNO_DEBUG=1` environment variable prints some tracing information I have manually added using the `notify!` macro. 

![image](https://github.com/kaleidawave/ezno/assets/26967284/23cad05d-cf07-4de7-b7a2-5c8ce8ea3b05)

I can see that subtyping is run 4 times for the above snippet. Each time checking whether structure generic (SG) on Array with `T (TypeId(11)) = number (TypeId(4)` is a **super**type of literal number type 3.

I also see *Here* being logged from `src\types\subtyping.rs:551` directly after. Hmm think I have found the culprit

https://github.com/kaleidawave/ezno/blob/24e35b98da5a6c705be3cce7eb56dc17bfcb32a4/checker/src/types/subtyping.rs#L551-L552

I think when I added that code I wasn't quite sure what should happen on that branch and so just default to saying it was fine (you have found a good case where it definitely is not). 

Just modified it to return `SubTypeResult::IsNotSubType(NonEqualityReason::Mismatch)` instead (as well as on the line a few above) and still passing all the tests. Will put in a PR to fix tomorrow!

[...]

PR #135, opened 2024-04-16
Fixes to things

kaleidawave:
- Flatten some of the files around poly_types / generics
- Add test for JSX
- Add tests for optional parameters (default value is undefined)
- Add test for variable and type shadowing
- Add some more Console methods
- Fix array subtyping issues #134 
- Add class method lifting for rest parameters
- Fix JSX children and template literal array length
- Wrap contributors and sponsors lines
- Update some dependencies
- Fix name in github-release.yml
- Fix `ezno check` to print warnings (fixed where it only printed diagnostics if there was errors)
- Added test for type mappings
- Less panics in partial syntax (+ new test)
- Subtyping example
- More `RangeMap` methods
- Rename `utils` to `utilities` (in checker)

Couldn't get the to the bottom of diagnostic emit overhead. Leaving for later


kaleidawave:
Now not cloning sources and reusing the `&str` source when printing diagnostics. But hasn't really touched the dial. 

```shell
 Summary
    Warning: Ignoring non-zero exit code.
    ./target/release/ezno check demo.ts --count-diagnostics ran
      1.05 ± 0.07 times faster than ./target/release/ezno check demo.ts --compact-diagnostics
      2.40 ± 0.12 times faster than ./target/release/ezno check demo.ts
```

Moving on
- Printing ~160 issues isn't a common case
- Can use `--count-diagnostics` or `--compact-diagnostics` in those cases
- Might be overhead in `codespan_reporting` library or just the fact that printing a lot of lines to the console is expensive

kaleidawave:
On the ongoing quest to figure why the diagnostics emit takes 60% of `ezno check` time on the total specification
- Trying replacing dynamic dispatch with static dispatch: https://github.com/kaleidawave/codespan/commit/ba6851bbcac82f85e0d103bfffa89142cc66fac5
- Results don't show a clear win 🤷‍♂️ https://github.com/kaleidawave/benchmarks/actions/runs/8751617511

Going to try on this branch later, maybe trying more complex diagnostics will show a change?

Issue #39, opened 2023-06-19
Add setters


kaleidawave:
At the time of writing Ezno supports getters on objects and classes.

It does this by treating getters the same as regular functions. Then when 'getting' the property it **evaluates the getter function**

https://github.com/kaleidawave/ezno/blob/502126217170121cadb12cad3debe1ff388a37f1/checker/src/types/properties.rs#L271-L292

This maintains all the behaviour of the event system in detecting side effects. etc

At the time of writing the setter part is not yet implemented. It should be the exact same code as 'getting' but calling the setter this time AND the setter should be called with one argument, the assigned value! **and return the assigner as setters (confusingly) ignore the returned value and just use the RHS value**. Similar to getter logic, the value of `this` should be `on` (the type the property is being looked up on). To handle the restriction, the `Result::Err` from `.call` should be pulled apart and any parameter mismatches should be turned into assignment diagnostics (can ignore the other kinds of errors for now)

At the end, this should type check
```ts
let a = 2;
const obj = {
   set value(v) {
       a = v;
   }
}

let b: 80 = (obj.value = 80);
let c: 80 = a;
```

Issue #37, opened 2023-06-19
Add positions to events


kaleidawave:
When a function returned value (which is looked up via events) doesn't match it error. This error doesn't contain the location of the return.

Many of the `Event`s could benefit from having a position (using `SourceWithSpan`) to mark where they occurred. 

To do this add a field with the type `SourceWithSpan` to an event and modify the synthesis and other code to make it compile. Then repeat for every variant of `Event`

kaleidawave:
Hey @NitantP, thanks for the interest. Hopefully the following answers your questions

1. So currently the `position` field in `ReturnedTypeDoesNotMatch` just points to the return type annotation. What would be ideal is for that it to **also** contain the positions of the *offending* `return` statements. Getting these positions for these errors would be really simple if they were held on return events. This sort of thing is already implemented for `AssignmentError::DoesNotMeetConstraint`
https://github.com/kaleidawave/ezno/blob/7cf648d5f83d08e9782ac8eb721b181d8aa2427b/checker/src/context/mod.rs#L1390-L1394 so when it emits an error, it also adds a label (additional information) that outlines where the variable type annotation is
 https://github.com/kaleidawave/ezno/blob/7cf648d5f83d08e9782ac8eb721b181d8aa2427b/checker/src/diagnostics.rs#L377-L392

2. and 3. Can't think of any more *metadata* that is needed on `Event` rn, so just a `Span` on each should be fine. I see that [SetsVariable](https://github.com/kaleidawave/ezno/blob/main/checker/src/events/mod.rs#L59) and [Throw](https://github.com/kaleidawave/ezno/blob/main/checker/src/events/mod.rs#L91) are currently the only two unnamed variants, feel free to make the named variants like the others!

I am hoping it shouldn't be too difficult to start on, I am happy to help with any problems that arise if you decide to give it go! **Note: you do not have to fully connect the errors up, I am thinking it might need another change for it to work perfectly.** But if all you could get events setup with a position for each then it would be really useful for now and future ideas!

kaleidawave:
Just a heads up @NitantP. #61 changes the structure of positions a lot so might be worth waiting for that to merge before starting

NitantP:
Understood - thanks, @kaleidawave!

kaleidawave:
Hi @NitantP #61 has been merged. So the repo code should be in a good state to work on this. Some updates:
[...]

PR #138, opened 2024-05-08
implement setters support as described in #39

PatrickLaflamme:
This PR implements setters support as described in #39. 

Note: In `run_setter_on_object`, we do an expect on the optional `setter_position` field, as the FunctionType.call method requires a SpanWithSource. I looked into refactoring to change the `setter_position` from an Option but it looked like it'd have wide consequences. @kaleidawave - would love your thoughts on this one.

### DIFF
--- a/checker/specification/specification.md
+++ b/checker/specification/specification.md
@@ -2894,3 +2894,37 @@ register(document.title)
 ```
 
 - Argument of type string is not assignable to parameter of type Literal\<string\>
+
+### Setters
+
+#### Setters invocation invalid assignment
+
+```ts
+let a = 2;
+const obj = {
+   set value(v) {
+       a = v;
+   }
+}
+
+let b: 80 = (obj.value = "some value");
+let c: 80 = a;
+```
+
+- Type "some value" is not assignable to type 80
+- Type "some value" is not assignable to type 80
+
+#### Setters
+
+```ts
+let a = 2;
+const obj = {
+   set value(v) {
+       a = v;
+   }
+}
+
+let b: 80 = (obj.value = 80);
+let c: 80 = a;
+```
+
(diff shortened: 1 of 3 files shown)


kaleidawave:
Awesome! Yep I believe the event positions were added in #37. There were some difficult cases so just decided it to make it Option and use None in those cases. If it's too difficult at the moment then fine to skip. 

Tests look good and are passing. Will double check and merge when I back week after next!

PatrickLaflamme:
Makes sense. Simulating the function call currently does not accept an Option for the SpanWithSource. I chose to fail closed here, panicking if a setter had no SpanWithSource because failing open would mean we would silently fail to evaluate the side effects of the setter function. 

If I understood your comment correctly, are you suggesting we update the FunctionType.call to accept an Option<SpanWithSource>?

kaleidawave:
> If I understood your comment correctly, are you suggesting we update the FunctionType.call to accept an Option<SpanWithSource>?

Actually the opposite, I think that `Event::Setter`s position field should be `SpanWithSource`. I think at the moment there are a few creations that don't have access to a position (but could!) such as `Environment::delete_property` and some of the class field registration. Those kind of set events won't trigger a setter so I think the `setter_position.expect("Setter position is required!");` is fine for the moment (and the tests still run). If you would like to improve that so that every event has a SpanWithSource position then let me know and I can look into it further and write up a issue explaining the cases :)

kaleidawave:
As for the rest, PR is great! Thanks for the contribution + adding tests for it!

kaleidawave:
Hey, again just added some additional tests. One thing that was missing in the PR is reporting the assignment type mismatch with the assigned value. It was being checked in `.call` (by regular function parameter checking), just wasn't being surfaced as the result was stored into `let _ =`. Not to worry have added some basic typing report it. 

https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#setter-assignment-type

https://github.com/kaleidawave/ezno/blob/c4ada6d3e52e0fad5a4eadcc513b3bfd31d2d0ed/checker/src/types/properties.rs#L743

Issue #42, opened 2023-06-27
Excess property checks (including the effects of spreads)


HiperMaximus:
It would be nice to allow Ezno to have more guards and checks than Typescript. One of those is performing excess property checks on spreading (which Typescript doesn't do), link to the issue: https://github.com/microsoft/TypeScript/issues/39998

I'm pretty sure there are other common footguns that Typescript has which would be nice to avoid here with a more strict typing. Great work!

kaleidawave:
This is a good suggestion. I haven't implemented excess property checks or spreading into a object yet but when I get around to it I will have a look at catching this!

kaleidawave:
I think this can be implemented here by raising a warning if `expected_type != TypeId::ANY_TYPE` and if the expected type does not have a property (aka the get_property == None)

https://github.com/kaleidawave/ezno/blob/24fb424f668df8fafb3e66ba76850ef6261e4b20/checker/src/synthesis/expressions.rs#L1060-L1069

PatrickLaflamme:
took a look at implementing this. Looks like this approach would result in new excess property warnings in these test cases:

* [Objects Checks](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#objects-checks)
* [Getters](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#getters)
* [Assignment](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#assignment)

In all three cases there is technically an excess property on the assignment, but curious to get your thoughts on whether that's actually what's desired here.

kaleidawave:
Awesome! Thanks for taking a look @PatrickLaflamme 

Yes I understand that [Objects Checks](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#objects-checks) and [Assignment](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#assignment) will raise new warnings for excess properties. Once this feature is added you can add a new bullet point to the markdown.

Unsure about the [Getters](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#getters) one though, I don’t an excess property there? If you have it working then you can put in a draft PR and I may be able to understand that one :)

PatrickLaflamme:
My mistake, it's actually [Generic Extends](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#generic-extends), not Getters. 

Agree getters would be a concern.

PR #139, opened 2024-05-08
implement excess params checks as described in #42

PatrickLaflamme:
Implements excess params checks as described in #42. Updates to the various tests are included. Two new tests are also added to `specification.md`.


kaleidawave:
Awesome. Same as other PR looks good to merge (just needs one pedantic linter fix). Will merge when back!

kaleidawave:
Another great PR! Have just done a few tweaks to the tests
- Tweaked some tests that previously had excess properties to not to avoid is possible.
- Used `declare let condition: boolean` to get an unknown boolean rather than a known one so that it didn't expect and merged together
- Added excess ones for assignments and function arguments

Also changed the message to add quotations around the property key as `Excess property property was was provided, but is not a property of type T` looked funny 😅

kaleidawave:
Thanks for adding this! It is cool to see my *expected property* feature allows this to catch issues that [TypeScript doesn't catch](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAsiDyAjAVhAxsKBeKBvKAZgPZEBcUAdgK4C2iEATgNxSICGDA-OdXY1AF8AsAChRaIhQDOmROThJUGbHlFR1hEuQCMAGjUaAdMdwGNG9g3IAmfSPPmAlhQBubADaOAJuQDMd8wE7ASA). Also good that it is a warning as it doesn't break any code. I don't know why the TypeScript team treats it as an error?

Going to showcase this addition in an upcoming blog post. Let me know if you want to include anything in that.

Also if you have any suggestions to improve the codebase, things that are confusing or not documented well let me know as I am going to do some refactoring soon!

PatrickLaflamme:
Will be diving into it again this weekend. Going to take a stab at the Proxy support ticket.

kaleidawave:
Hey, added some additional tests for this: https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#excess-property. Because of the expected type / forward inference thing this also works for returns.

(also changed the excess property lookup to use the existing property lookup, so should be a little faster).

PR #142, opened 2024-05-17
Updates to the parser and CLI

kaleidawave:
- Fixed array destruction syntax (and added support to checker)
- Update CLI things
- Added retaining blank lines for formatting
- Use BufferedStandardStream
- Updated release compiler options
- Added missing `run` checker example
- Optional class properties
- Removed Eq from parser items
- More additional syntax for destructuring with type annotations
- Added setter assignment subtype check and more setter tests
- Add more excess property tests


kaleidawave:
image

🖕

kaleidawave:
A combination of retaining blanks lines as and the bizarre `;` empty statement causes the fuzzing to break the `parse -> print -> parse -> print` equality. As this is an edge case and doesn't break anything I am going to merge and have another go at fixing it once I have a better idea of how to solve it.

PR #143, opened 2024-05-19
add `bacon.toml` for easier new dev experience

PatrickLaflamme:
Proposal to use [Bacon](https://github.com/Canop/bacon) to make the standard operations of running the specification tests a little simpler during development.

How to use:

1. install `bacon` with `cargo install --locked bacon`
2. press `1` to run only the `specification.md` tests once.
3. press `2` to run only the `staging.md` tests once.
4. press `3` to run all tests once.
5. press `4` to run only the `specification.md` tests and watch for changes.
6. press `5` to run only the `staging.md` tests and watch for changes.
7. press `6` to run all tests and watch for changes.

Bacon will watch for changes in the source sets and re-run the tests on change. This can help optimize the feedback cycle and accelerate development.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,6 +60,37 @@ cargo run -p ezno-parser --example parse path/to/file.ts
 cargo run -p ezno-parser --example lex path/to/file.ts
 ```
 
+### Bacon (script runner)
+
+The [Bacon script runner](https://dystroy.org/bacon/) is configured for this repo. This can watch your files and re-run things like checks or tests on file change.
+The configuration is managed in the [`bacon.toml`](./bacon.toml) file. The configuration has dedicated jobs for the checker specification tests mentioned above. 
+
+#### Installing Bacon
+
+To install bacon, simply run `cargo install --locked bacon`, and you're ready to go.
+
+#### Using Bacon
+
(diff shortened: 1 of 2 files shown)


kaleidawave:
Awesome, just looked into Bacon. Looks like it solves a problem of unifying the current jumble of scripts this project uses.

Few questions before I merge
- Could you add a short overview of how to use it (similar to the PR body) to `CONTRIBUTING.md` under a heading like `### Bacon (script runner)`?
- I would quite like to use this, but are a little bit unsure of the watch mode (as similar watch tools that run compilation lots kill my laptop battery). Does it run ALL the scripts in the background or just the one selected? Is there a non-watch mode?

Other than that looks great, thanks for introducing me to it (hadn't heard of it before). Also might add some more scripts that I use to it later.

PatrickLaflamme:
added some details in the `CONTRIBUTING.md` file. Also created some jobs that don't watch for changes. If you want to re-run, just press the hot key to trigger a run again. 

I find it particularly useful in my IDE to have a terminal open running the spec tests, and then as I make changes I can see how my changes affect the test outcomes. Makes iterations SUPER fast.

kaleidawave:
Thanks for adding documentation to contributing. Excited to try this!

Issue #81, opened 2023-11-13
Multiple files as entry point


kaleidawave:
Update: after #102 have added multiple entry points to the checker. However it is not fully wired up to Ezno's CLI with:

https://github.com/kaleidawave/ezno/blob/1aa77e2aabb9533fb4fe90fa1ac9aaa24136256b/src/cli.rs#L170-L172

The next part (that you can contribute to):
- Get a `Vec<PathBuf>` from a input path.
  - This could be a glob pattern or just a comma separated list
  - Should consider whether this method will work with the [notify crate](https://docs.rs/notify/latest/notify/). While `check --watch` isn't currently implemented, it would be good if this addition can work with a watch mode in the future

> Note I have not checked side-effect edge cases in #102
--- 
Currently there is only one entry path. OP wants to check all files in the directory. Note each entry point also type checks its imports, so this isn't necessary for checking imports and exports. 

![image](https://github.com/kaleidawave/ezno/assets/26967284/6b30ef5a-86f9-4ffb-96aa-af8ef702a197)

Rather than running `entry_points.for_each(|path| check_project(path, ...))` this needs to be built into the internal `ModuleData`. So it should be callable as `check_project(vec![...])`. 

Things to think about
- Side effects in modules (want to reuse some existing data, but make it distinct)
- How to get a `Vec<PathBuf>` from the *glob* passed to the CLI
    - How this works in the CLI watch mode

### Discussed in https://github.com/kaleidawave/ezno/discussions/80

<div type='discussions-op-text'>

<sup>Originally posted by **o-az** November 11, 2023</sup>
How can I run `oxidation-compiler` to typecheck multiple files by passing a glob string / directory path?

Works:
```ts
npx oxidation-compiler@latest check src/file.tsx
```
None of these commands check all files:
```ts
npx oxidation-compiler@latest check src/*.tsx
```
```ts
npx oxidation-compiler@latest check src/
```
```ts
npx oxidation-compiler@latest check src/*
```
![CleanShot 2023-11-11 at 10 48 53@2x](https://github.com/kaleidawave/ezno/assets/23618431/be44ae4c-5687-4f07-8af1-32807191aec2)</div>

kaleidawave:
Implemented in #155

PR #145, opened 2024-05-21
Parser and CLI updates

kaleidawave:
- Add asserts type (technically function return type)
- Add `as` for inteface rule + more
- Add binary *self-upgrade* system
- Start on #81 with allowing multiple entry points
- Rename some examples
- Add 'pretty-printing' example


kaleidawave (commented):
Some things still to do:
- Finish + test the multiple file entry thing
- Test over the self-upgrade system
- Look at the fuzzing issue again
- Maybe more pretty printing for chaining and stuff

Issue #136, opened 2024-04-18
`never` type lookup - "Cannot find type never"


CharlesTaylor7:
```ts
function loops(): never {
    while (true);
}

loops() satisfies never
```

When I check this block of code I get an error of  "Cannot find type never".

I suspect this is an issue at the parsing stage, because there is already code in the subtyping logic that checks for the `never` type.

CharlesTaylor7:
@kaleidawave I started a branch for `never` related issues. It's necessary to get `Pick`, `Omit`, `Exclude` and `Extract` utility types to work properly.
https://github.com/CharlesTaylor7/ezno/tree/never-type

Also, confirmed that tsc does enough reachability analysis to verify the return type of an infinite loop is `never`, instead of `void`.
https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAGznADgZwBQEoBciYApgG7EBOiA3gFCIOIDuAFjMsYtlBSMbgG5aAX1q1UGHLkSYAhrEzAYxTETKVaQA

CharlesTaylor7:
@kaleidawave It's an interesting question you posed about shadowing `never` with another type. I've never tried to do that before, so I checked in TS playground. Apparently certain types are treated like keywords and cannot be shadowed.  

https://www.typescriptlang.org/play?#code/C4TwDgpgBAdhBuEBOUC8UCsBuAUAMwFcYBjYASwHsYAeORJAPgAo8KKAuWBZASigG8cOAL5DQkKAGdgSMjADmaKABZchEuSrVpshc1YcpMufL6CRQA

kaleidawave:
Interesting. That's actually quite a good idea. I could see having a type parameter named `never` could cause a lot of a confusion. 

So your addition to the type annotation name lookup is correct and my suggestion about the `Root` context named typed would create diverging behaviour with TSC. 

I think type shadowing works ATM (although it seems I have added any tests for it). Adding prevention and emitting of errors around invalid type aliases and type parameter names needs adding.

Issue #34, opened 2023-06-16
Type of arrays and other collection types


kaleidawave:
Given an array like
```ts
let x = [1, 4, "item"]
```
It is typed (or will be as `{ [0]: 1, [1]: 4, [2]: "item" }`). The problem is how to reference its prototype. 

Array is a generic structure so it could be `Array<1 | 4 | "item">`. However it would have to be modified every push...

It is needed to test equality
```ts
let x: Array<number> = [1, 4, "item"]
```

Alternatively the `T` type could be figured out from items. Along the lines of `interface Array<T is this[number]>`...

This also affects `Set`, `Map`...

kaleidawave:
Won't go into too much detail here but
- Lookup generics have solved type parameters of collection types
- User created `Promise` is a `Type::SpecialObject` and will get a special implementation

Still to implement before closing
- Lookup generics for Set and Map

PatrickLaflamme:
@kaleidawave - Is this covered in your currently open PRs? If not I'd like to take a stab at it.

kaleidawave:
Yeah I hopefully will get `Set`, `Map` and `Promise` type structure working in #146. After that this issue should be finished

kaleidawave:
Didn't happen in #146.

There exists a mechanism called "lookup/computed generics". It [works for subtyping](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#array-property-checking) and [other places](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#computed-generics-from-collection). 

Lookup generics is manually set, it would be good if it could infer for any array (via trying to find a member, preferable contravariant else covariant for some edge cases).

Will test for `Map` & `Set` soon.

> Proxy is a different issue #33 and Promises are different #168

PR #146, opened 2024-05-24
Internal improvements to subtyping and substitution

kaleidawave:
- Change subtyping to use single state structure (rather than generic structures)
- Change substitution to use single immutable object for arguments
- Changes to accommodate the above
- `* extends * ? `, `infer U` and distributive extends
- Add `never` #136 + or folding
- Inline `get_fact_by_type` into the only place it is used: `get_property_unbound`
- Start of some try-catch stuff
- Start of inference stuff (fully won't be in this PR)
- Start of proxy stuff (fully won't be in this PR)
- Changed `this` object creation as an event (move logic internal to function calling)
- Allow for `Func.prototype.x` thing
- Flattened events
- Unknown-ness after for loops now uses closure behaviour


kaleidawave:
Wondering whether to make events flatten and reuse them more (instead of state and `ApplicationResult` which sorts of duplicates information).
- This would be the state of generators and promises around nested a lot easier (just one dimensional index to record)
- Would make it easier to run *half branches*
- Would need to change if and for to be next

Also on the topic of events: should have an option to record all set and variable events for usage in extra diagnostics

kaleidawave:
image

Still to do
- Way to disable the simple.d.ts during tests
- Get conditional and iteration events working again
  - Get array working
  - Get more exception things working
  - Something around free variable inversion
  - Does application need rest of events (unfortunately this can't be referenced contiguously during nesting so need to do something like generic chain)
- Work on more mapped types
  - Maybe get typeof and keyof working at the same type
- Rethink lookup generics. If we add `typeof` and `keyof` it might be a better way than the current #34 
  
Done
- `this` setting improvements (not as an event anymore which is a little simpler)
- Class super call works (sort of)
- Functions are now hoisted

kaleidawave:
For mapped types: could
1. Type arguments to enum TypeArgument = Explicit | PropertyKey | ...
2. Specialisation arguments have a property key field which is can use to specialise a mapped type?

kaleidawave:
Managed to squeeze some new features in here, bringing the passing tests to 235!

kaleidawave:
This is weird. But we move
```ts
error: 
    ┌─ demo.ts:414:7
    │
410 │     function getA<T extends { a: strinthread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/source-map-0.15.0/src/filesystem.rs:141:24:
index out of bounds: the len is 2 but the index is 18446744073709551615
g }>(p: T) {
    │                                            ---- Parameter has type T
    ·
414 │     getA({ a: 2 })
    │ 
```

PR #147, opened 2024-05-24
Parser improvements

kaleidawave:
Really just want to see the full fuzzing results to fix that issue

Also want to implement
- Pretty function parameter printing


kaleidawave:
Probably some other things could be worked on in this PR but we move

PR #151, opened 2024-05-31
Fix getting started link in README

jasikpark:
Updates a link to the checker getting started docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 A JavaScript compiler and TypeScript checker written in Rust with a focus on static analysis and runtime performance.
 
 > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


kaleidawave:
Whoops. I wonder if there is a simple (no config) markdown checking (formatting and no broken link) tool that could be [added in the CI](https://github.com/kaleidawave/ezno/blob/main/.github/workflows/rust.yml#L47-L61) to prevent it happening the future?

jasikpark:
I wonder if https://lychee.cli.rs/recipes/relative-links/ is the right choice? It looks like it supports local relative links and it's written in rust!

kaleidawave:
Yeah will have a look. I use the https://github.com/DavidAnson/markdownlint vscode extension for some things but I don't like the defaults. Need something in the middle?

PR #156, opened 2024-06-03
Add basic narrowing

kaleidawave:
Adds narrowing of variable reference types based on JavaScript equality constructs.

Currently adds the following
- `===` narrowing
- `typeof` narrowing
- Support for `&&` and `||`
- De Morgans's law edge cases
- `assert *value* is *type*` checking
- Some number extras which require additions from #186
- `(X | Y).tag === ...`
- `*key* in obj` case removal and creation
- `instanceof`

Support across variable declarations and functions thanks to more dependent types

> Additionally the `LocalInformation` merging now handles final events better. (see new control flow test, which would previously be `3 | 7`) AND AnonomousObjectLiteral stores the properties on the type


kaleidawave:
Some things to do:
- Backing id changes
- Ands (should be fine using the current system)
- Or
  - Needs to buffer results and combine in final step?

kaleidawave:
Still to do:
- Think about filtering objects by predicates on properties. Aka `{ tag: 1, d: string } | { tag: 2, d: boolean }`, after `obj.tag === 1` becomes `{ tag: 1, d: string }`
  - InformationChain is already accessible here. Narrowing needs to pass down this information
- Whether variable reference narrowing is enough?

checker/src/features/narrowing.rs
```diff
@@ -0,0 +1,470 @@
...
+						&mut items,
+						information,
+						types,
+					);
+					types.new_or_type_from_iterator(items)
+				};
+				into.insert(on, types.new_narrowed(on, narrowed_to));
+			}
+			constructor => {
+				if let Some((lhs, rhs)) = as_logical_and(constructor, types) {
```
kaleidawave, line 244: @danvk thought you might find this interesting! I watched the [Michigan TypeScript](https://www.youtube.com/@MichiganTypeScript) video and read your blog post on [implementing inferred type predicates](https://effectivetypescript.com/2024/04/16/inferring-a-type-predicate/). (congrats on getting it merged, it is a great feature!)

In Ezno it was a lot easier to add narrowing with its more *dependent* type system. In Ezno the expression `typeof x === "number"` has a type like `EqOperator<TypeOfOperator<typeof x>, "number">` which makes getting the information for narrowing pretty simple (you can sort of see that in the lines above). So Ezno has something closer to your original intuition rather than the TSC flow node control flow structure.

One of the interesting parts was "Pathological Cases and an Insight" with this case edge case in the `else`
```ts
function flakyIsStringRewrite(x: string | null) {
  if (typeof x === 'string' && Math.random() > 0.5) {
    x; // type is string
  } else {
    x; // type is string | null
  }
}
```
this helped me fix a bug in my implementation. For else branches I run `get_narrowed_values(condition, true)`, where `true` negates the condition in the `if`. If I have the `&&` expression above with `negate = true` it then means that it applies one of the De Morgans law. So it now treats `!(a && b)` in the code as `!a || !b` and it only adds a *narrow* if both sides have a condition for the same variable (=`on` here) (the zipping code being `for (on, lhs_request) in lhs_requests {	if let Some(rhs_request) = rhs_requests.get(&on) { ...`). 

[I think I have got this covered in the tests here](https://github.com/kaleidawave/ezno/blob/narrowing/checker/specification/staging.md#edge-case)

Don't know if there is an equivalent section in the TS narrowing logic but thought my intuition thinking about De Morgan's laws and generating `A | B` only if both sides of an `||` condition give narrowed values for the variable was a nice way to handle this case.

kaleidawave:
Most definitely missing some functionality but the basics are in so merging 🎉

Issue #35, opened 2023-06-19
Inferred generic (poly type) constraints


kaleidawave:
A aim of Ezno is to work without writing type annotations. Variables don't need them because of the value system, return types are based on the body. The final piece is the unknown-ness of inputs to functions. 

This can be in several places:

#### Explicit parameter
```ts
function func1(a) {
    return a.x // The constraint of `a` should be inferred as `{ x: any }`
}
```

#### Non restricted variable
```ts
let a = 4;
function func2() {
    return Math.sin(a) // The constraint of closed over references `a` should be inferred as number (with strict casts at least)
}
```

#### Cycles (recursion)
```ts
interface Node {
    parent: Node | null,
    x: 
}

function depth(obj: Node) {
    if (obj.parent) {
        // `depth` hasn't been checked yet. A restriction should be made
        return depth(obj.parent) + 1
    } else {
        return 0
    }
}
```
See #29

#### Others
- Throw...

### `any` as `unknown`

To better support existing codebases `any` should be treated as `unknown`. No annotation should be treated as `unknown` 

### Combining restrictions
When two restrictions are made they are combined using `&`
```ts
interface X {
    readable: boolean
}

interface Y {
    name: string
}

function x(p: X) {}
function y(p: Y) {}

function z(param1) {
    x(param1)
    y(param1)
}
```

`param1` is now restricted to `X & Y`. 

*this might create impossible types (string & number), which is fine but a warning should be raised*

### Places where constraint inference is done
- Calling a type
- Getting a property
- subtype-ing. If `x` has a unknown restriction and `x <: number` is evaluated, then `x`'s restriction becomes `number`

---
I had up to here existing implemented, the hard part is now:

### Nesting
```ts
function x(obj) {
     return obj.prop1.prop2.prop3
}
```
This needs to place a inference condition on a inference restriction 😅

### The non-local `unknown` issue
If `unknown` is found on a fixed restriction, then a nested dynamic restriction might have to be on a fixed restriction.

```ts
interface Something {
    prop: any
}

function doSomethingWithSomething(s: Something) {
    return Math.sin(s.prop)
}
```

*extended from https://github.com/kaleidawave/ezno/blob/main/checker/docs/inference.md*

It is still in progress the best way to tackle this, interested if people have real world parameter inference examples they want supported:

kaleidawave:
Thinking of some problems with mutations. For example given the following short example

```ts
function func(prop, callback) {
    const { a } = prop
    callback(prop)
[...]

PR #157, opened 2024-06-03
Improvements to properties + intrinsic types

kaleidawave:
300 passing 🎉

- Replace `todo!` under /synthesis for fixes or unimplemented error (for #152)
- Under option (`true` by default): if `for (let i = 0; ...)` or similar where the initialiser is number like. Assume `i` has constraint of number (assigning to any other type is not an advisable pattern). Assuming this constraint would skip work that will be done when #35 is added!
- Added TSC intrinsic string types
- TSC `Uppercase`, `Lowercase` intrinsic string types
- Added TSC `NoInfer`
- Ezno `CaseInsenstive`
- Added some WIP Ezno intrinsic number types
- Added some WIP `Not` type and `Exclusive` type
- Added new slice subtyping mechanism
- Have got the mapped property key of mapped types partially working. 
  - Including modifiers
  - `GenericChain::Mapped`
  - Access is pretty good
  - Subtyping is still a bit WIP
- Have got `delete` checking in
  - No test for doing inside of function
  - Breaks the `pop` synthesis, so removing for now
  - As effect broken :(
- `in` is handled a bit better across functions
  - Getting pinned down to decidable is quite difficult
- Understanding break *works*
  - Turns out that was being run unknown as was trying to run 460 events on top level (default limit is 300)
  - Got it down to 360 by not queueing some `SetVariables` where it is in the scope up during application
  - And further down by changing the values in the test
  - However now the break isn't being fired. Might be to do with the carry, might not be ?
- Delete as effect
- `Object.defineProperty` and `Object.getOwnPropertyDescriptor` (constant functions)
  - `PropertyValue::Configured`
    - Might be bad for performance for `a.push` but oh well

- Sorting for numerical properties
- Sorting for later assignments (removed old weird `.reverse` thing)
- Proxy
	- Added `get` and `set` for `Proxy`
	- Added *basic* subtyping for `Proxy` (probably doesn't work in more general cases)
- Added `Object.assign`
- Added `configurable` checks (change + delete)
- Array.prototype.map
- Register as effect
- Merge definition hoisting into regular hoisting
- Cyclic alias (+ improved cycle detection and checking here)

> `StringBuild` test seems a bit temperamental :/


kaleidawave:
256 🎉 

Going to remove some WIP things (operator checks, failing events, `.d.ts` emit) and split in to separate PRs. Also the template literal strings array needs changing (but should do that in the parser first , could do in #173)

Before merging
- Fix `.map` not calling error?
  - Also computed generics not working here
- Have a look at mapped type issues (Including the new `GenericChainLink` variant)
- Fix the string slice extends infer thing (the mapped property as works so not sure?)
- checking_with_function_prototype

kaleidawave:
269 🎉 
```shell
failures:
    specification::collections::array_map
    specification::forward_inference::computed_generics_from_collection
    specification::staging::calling_type::array_slice_matching_pattern
    specification::staging::calling_type::properties_on_big_or
    specification::staging::calling_type::regexp_patterns
    specification::staging::exceptions_and_try_catch_finally::throw_through_internal_callback
    specification::staging::extras::exclusive
    specification::staging::mapped_types::as_rewrite
    specification::staging::mapped_types::negated_optionality
    specification::staging::mapped_types::negated_readonly
    specification::staging::mapped_types::optional
    specification::staging::mapped_types::specialisation

test result: FAILED. 269 passed; 12 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.05s
```
Hopefully can fix `.map` and make some progress on the others 🙏

kaleidawave:
For mapped types: the last thing to fix is subtyping based on the mapped type being on the left. The problem to solve is indexing a type based on a type key. This sort of works ATM for `Array[number]` but not much more. I have just added support for reading on `ty["a" | "b"]` eagerly in `get_property_unbound` via splitting the branch and building a `Logical::Or`. 

For more complex types: There exists `Logical::BasedOnKey` for a value based on the RHS key. I think this should be extended to somehow RHS. Somehow the value should be based off `get_properties` with a filter?

Aside: this is duplicate of lookup generics get_key_on...? Will come back to this maybe with default generics, or covariant lookup.

Ahh its all really complicated and making this PR take a long time. But when it is done :chefskiss:

---

Once that works then will have the proper foundations for `.map` (which uses properties)

kaleidawave:
300 🎉 

image

Issue #154, opened 2024-06-01
Panic when parsing let declaration with no declarations


jasikpark:
```
!!sqve+l//
















l
let
///
oOo�str!er
```

as input for the `module_roundtrip_naive` fuzzer produces this error:

```
Running: fuzz/artifacts/module_roundtrip_naive/crash-cbb815ba074710aad65fc66c094f2e1600e0c80e
thread 'AST parsing' panicked at /Users/calebjasik-personal/Git/ezno/parser/src/declarations/variable.rs:274:63:
called `Option::unwrap()` on a `None` value
```

https://github.com/kaleidawave/ezno/blob/34eecf1198229424560bf390fd87b0ec82d407db/parser/src/declarations/variable.rs#L273-L276

jasikpark:
Seems like there needs to be an error thrown when there's a `let` but no associated declaration?

kaleidawave:
Hmm yes. Could start with `let mut declarations = vec![VariableDeclarationItem::from_reader(...)]` but yes emitting a `ParseError` if `declarations.is_empty` would be better because it could be allowed under `partial_syntax` so that [this works](https://kaleidawave.github.io/ezno/playground/?code=DYUwLgBAUAxg9gOwM6QB4QLwQEwG4g)

Will investigate and fix!

jasikpark:
I'm thinking I may do work to integrate these mature fuzzers with OSS-Fuzz, where they can be run for longer and catch bugs like these, since it seems like they aren't catching anything on 5m Github Action runs anymore, right?

kaleidawave:
There are things being caught! I am unsure how this example hadn't been flagged by CI yet? Maybe just a rare state. 

ATM things in the type checker have a lot higher priority so I am quite happy with the current level of parsing fuzzing. If you want to and is simple to change/upgrade sure! I do want to keep it under 2 mins though, it is a bit annoying to wait any longer for results.

kaleidawave:
Fixed in https://github.com/kaleidawave/ezno/pull/158

PR #158, opened 2024-06-03
Few parser fixes

kaleidawave:
- Fix capitalisation in `ForLoopStatementInitialiser`
- Fix for #154
- Fix for #161
- Fix for `parser/examples/code_blocks_to_script.rs` and `.github/workflows/performance-and-size.yml`
- Raise parse errors for *trailing* syntax in variable field currently not raising a parse error
- Add different array printing ([see comment below](https://github.com/kaleidawave/ezno/pull/158#issuecomment-2169621017))

### DIFF
--- a/.github/workflows/clippy-rustfmt-fix.yml
+++ b/.github/workflows/clippy-rustfmt-fix.yml
@@ -1,0 +1,37 @@
+name: Run automated fixes on current branch
+
+on:
+  workflow_dispatch
+
+env:
+  CACHE_PATHS: |
+    ~/.cargo/bin/
+    ~/.cargo/registry/index/
+    ~/.cargo/registry/cache/
+    ~/.cargo/git/db/
+    target/
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/cache@v4
+        with:
+          path: ${{ env.CACHE_PATHS }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      
+      - name: Run automated fixes
+        run: |
+          cargo clippy --fix
+          cargo fmt
+
+      - name: Commit
+        run: |
+          git add .
+          git commit -m "Run clippy --fix & formatting"
+
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        
+          git push
--- a/.github/workflows/performance-and-size.yml
+++ b/.github/workflows/performance-and-size.yml
@@ -50,23 +50,23 @@ jobs:
         shell: bash
         run: |
           # Generate a file which contains everything that Ezno currently implements  
-          cargo run -p ezno-parser --example code_blocks_to_script ./checker/specification/specification.md ./demo.ts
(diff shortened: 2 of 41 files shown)


kaleidawave:
#### Pretty formating of array

Currently if some structures including array have check to see if their content printed on the same line overflows a specific count. This works well (see `parser/examples/pretty_printing.rs`). However in a project I am working on I had a case where I have a array of lots of literals.
```ts
const x = [1, 1, 0, 0, 1, 0, 1, ... *etc* ]
```
as it breaks the max-line width threshold it is pretty printed as
```ts
const x = [
  1, 
  1, 
  0, 
  0, 
  1, 
  0, 
  1,
   ... *etc* 
]
```
I wonder if in this case it could be alternatively done in half-way manner, where the items are distributed?
```ts
const x = [
  1, 1, 0, 0, ...
  1, 0, 1, 0, ...
   ... *etc* 
]
```

kaleidawave:
Help @addisoncrump @jasikpark.  Not sure why it is now requiring `x86_64-unknown-linux-musl`, when the rest of the tests run on `x86_64-unknown-linux-gnu`?

```
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-unknown-linux-musl` target may not be installed
  = help: consider downloading the target with `rustup target add x86_64-unknown-linux-musl`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`

For more information about this error, try `rustc --explain E0463`.
error: could not compile `libc` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
Error: failed to build fuzz script: RUSTFLAGS="-Cpasses=sancov-module -Cllvm-args=-sanitizer-coverage-level=4 -Cllvm-args=-sanitizer-coverage-inline-8bit-counters -Cllvm-args=-sanitizer-coverage-pc-table -Cllvm-args=-sanitizer-coverage-trace-compares --cfg fuzzing -Clink-dead-code -Cllvm-args=-sanitizer-coverage-stack-depth -Cdebug-assertions -C codegen-units=1" "cargo" "build" "--manifest-path" "/home/runner/work/ezno/ezno/parser/fuzz/Cargo.toml" "--target" "x86_64-unknown-linux-musl" "--release" "--config" "profile.release.debug=true" "--bin" "module_roundtrip_naive"
Error: Process completed with exit code 1.
```

It was working yesterday (maybe a change in 1.79 that has just come through, not sure though) and not sure whether changing the CI `dtolnay/rust-toolchain@stable` target is a good fix or would even work?

addisoncrump:
You seem to have run into this: https://github.com/rust-fuzz/cargo-fuzz/issues/355

I'm surprised that hasn't already been fixed... I bumped the upstream issue.

kaleidawave:
Need to look into this fuzzing issue before merge. Understand why broken, but not sure why now? There hasn't been an update to `cargo-fuzz` in a while? Manually setting the version hasn't seemed to work?

kaleidawave:
Installing from the GitHub repository works? (`cargo install --git https://github.com/rust-fuzz/cargo-fuzz.git`). Only adds 18s to the action (to build the binary from source)? Should be okay, until https://github.com/rust-fuzz/cargo-fuzz/pull/365 is merged!

lishaduck:
Are the issues linked resolved? You didn't put a colon after fix, so they didn't autoclose 😉

PR #163, opened 2024-06-13
Fix typos in README

lishaduck:
Add a few commas here and there and a period or two, remove some old comments.


kaleidawave:
Thanks!

PR #169, opened 2024-06-24
Playground improvements

kaleidawave:
- Add performance
- Write URL to clipboard
- Remove lz-string
- Fix for lots of text in playground
- Not default button style

### DIFF
--- a/src/playground/package.json
+++ b/src/playground/package.json
@@ -20,7 +20,6 @@
     "@codemirror/view": "^6.23.0",
     "@lezer/highlight": "^1.2.0",
     "@lezer/javascript": "^1.4.13",
-    "ezno": "=0.0.21",
-    "lz-string": "^1.5.0"
+    "ezno": "=0.0.21"
   }
 }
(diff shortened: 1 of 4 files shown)


src/playground/package.json
```diff
@@ -20,6 +20,6 @@
     "@codemirror/view": "^6.23.0",
     "@lezer/highlight": "^1.2.0",
     "@lezer/javascript": "^1.4.13",
-    "ezno": "=0.0.21"
+    "ezno": "^0.0.22"
```
jasikpark, line 23: it'd be great if the playground mentioned the version it's using 👀

kaleidawave: Check line 162 :)

kaleidawave:
Just published this to public from this branch: https://kaleidawave.github.io/ezno/playground/ (yeah not great that `crates.io+npm`, `github releases` and updating the playground has independent deploys. is flexible but might be a better way).

There is one more thing I might add later to the site. Before merging this.

PR #172, opened 2024-07-03
Log the error when we get an input that errored after being reprinted

jasikpark:
This PR updates `parser/fuzz/module_roundtrip_structured` and `parser/fuzz/module_roundtrip_naive` to log the error received when a snippet causes an error on the second parse after a reprinting


kaleidawave:
Thanks, keep reporting any new mismatches!

Issue #166, opened 2024-06-20
List possible variables and type names when cannot find reference


kaleidawave:
When a variable name cannot be found [an error is raised](https://github.com/kaleidawave/ezno/blob/main/checker/specification/specification.md#variable-references-does-not-exist)

For the current this means the `somethin` will raise a type error
```ts
const something = ...
console.log(somethin)
```

**However** the current error for this sort of not found reference error is:
```
error: 
    ┌─ demo.tsx:799:12
    │
799 │     const a = not_found
    │               ^^^^^^^^^ Could not find variable 'not_found' in scope
```
It would be better for the error to include possible alternatives that are in scope:
```
error: 
    ┌─ demo.tsx:799:12
    │
799 │     const a = not_found
    │               ^^^^^^^^^ Could not find variable 'not_found' in scope
    │     
    |     Did you mean `found` or `is_found`
```

---

I started on this, but didn't finish. You can add the `possible` fields to the `TypeCheckerError::CouldNotFindType`, use the `levenshtein` crate to find similar (aka low distance) names to the specified one and `parents_iter().map(|env| env.variables.keys())` to get variable names in the scope (+ `named_types` for types)

https://github.com/kaleidawave/ezno/blob/5191329edd448c364914917297bef2846527b91b/checker/src/diagnostics.rs#L293-L298

https://github.com/kaleidawave/ezno/blob/5191329edd448c364914917297bef2846527b91b/checker/Cargo.toml#L40

I would recommend approaching it by
- Make a public function in `lib.rs` that has the signature `pub fn get_closest(items: Iterator<Item=&str>, closest_to: &str) -> Option<&str>` (want it public as will be useful in the LSP #22)
- Wiring it up here https://github.com/kaleidawave/ezno/blob/bb03d1f9dc81c4f782113e1892653ef91e739cbf/checker/src/context/environment.rs#L892
- Creating a demo file with the above example
- Using `cargo run -p ezno-checker --example run_checker`  on the above example

vHugoObject:
Can I take this?

kaleidawave:
Sure @vHugoObject, no one has taken it. Have updated the description. LMK if you get stuck on anything!

vHugoObject:
To be clear, I should create a typescript file with some code where a variable cannot be found and place that example file in checker/examples so that it can be ran with cargo run -p ezno-checker --example run_checker?

kaleidawave:
[...]

PR #180, opened 2024-07-29
List possible variables, types, properties and imports when cannot find reference

vHugoObject:
For #166. A get_closest function was added using the levenshtein crate. This function is used to create a list of possible variables, types, properties and imports when a cannot find reference error is raised.  The specification tests have been updated to reflect the new error message. I would like to note that in the code, there are currently two functions that can be used to raise a cannot find reference error for properties: PropertyDoesNotExist and ExcessProperty. My pull request only makes changes to the PropertyDoesNotExist. ExcessProperty would need lifetime annotations for a "possibles" variables, which would have required a bigger revision of the TypeCheckWarning enum.

### DIFF
--- a/checker/specification/specification.md
+++ b/checker/specification/specification.md
@@ -49,7 +49,7 @@ a satisfies number
 const a = c
 ```
 
-- Could not find variable 'c' in scope
+- Could not find variable 'c' in scope. Did you mean a?
 
 #### Assigning before declaration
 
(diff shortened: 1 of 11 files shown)


kaleidawave:
Awesome! Thanks for the changes, all looks good. Especially for going the extra mile to add suggestions for properties and files.

Clippy error seems to be caused by new Rust update, so will ignore for now.

In the future may change the testing and diagnostics but won't fuss about that for now.

Issue #165, opened 2024-06-20
Cannot parse `<T>(t: T) => T` under JSX mode (problems with expression prefix)


kaleidawave:
The JSX lexing kicks in after the previous token is considered a *expression prefix* (this works for JSX lexing). So for the colon `:`, JSX lexing mode kicks in

```
let x = { a: <*JSX mode* ... }
let y = ... ? ... : <*JSX mode
```
> IMO the former should be `(:)=` and the latter should be `(if) ... then ... else ...` but 

This means that the following doesn't work:

```ts
function func(a: <T>(t: T) => T) {}
//             ^ ^^^
//               Thinks this is JSX because of the : before 
```

This also doesn't parse correctly
```ts
type X = (<T>(t: T) => T) | number;
//     ^
```

Not sure how to tackle this. The lexing is designed to be context-free. Communicating with the parser would require architectural changes. Maybe the lexer could hold a state so that it knows it is in some mode? Maybe could lookahead in the lexer and check that the next == '('?

It is a shame because the expression prefix mode works really well ATM and covers regular expression stuff. 

I think this works on non JSX parsing mode?

Alternatively this works ATM

```ts
function func(a: { b<T>(t: T): T }["b"]) {
  print_type(a)
}
```

---

This is not to do with generic arrow function syntax (don't do this)
```ts
const x = <T>(t: T) => t;
```
or the casting syntax (never do this)
```ts
const x = <string>b;
```

jasikpark:
Curious how esbuild parses it, since they obv support distinguishing between TSX types and elements declarations: https://github.com/evanw/esbuild/blob/fc37c2fa9de2ad77476a6d4a8f1516196b90187e/internal/js_parser/ts_parser_test.go#L556

Would be sweet to replicate these tests if possible too.

kaleidawave:
A few more problems with the current "expression prefix" token thing for changing the lexing mode:

- `for (const ... of /x/)` works but that breaks `x.of.y` because it thinks `.y` is a number literal without the `0` prefix. (also others)
- and `.return` is also a valid property access so `.return < 5` breaks in JSX mode

I think more 'state' in the lexer could solve this. There is current low level state, but a higher level state that captures the state here
- property access for `(?).return`
- `for (await) (... of`)
- *others*?

kaleidawave:
This should be fixed after #191

PR #184, opened 2024-08-07
Initial RegExp

### LINKED ISSUE #175: Implement RegExp/regular expression/regex analysis
For RegExp objects, the following should be tracked
- Valid syntax
  - Duplicate capture groups
- Whether it has been *used*
- Possible capture groups for `String.prototype.match`
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#instance_properties
   - Some can be implemented by `this.flags.contains("*")`

This can be tracked for simple literals. For the constructor this can be done using a constant function with an decorator.

For the best checking I think it would be good to use an external library. (this could be under a flag and only used for testing and release for debug builds)

### LINKED ISSUE #183: Add regex checking
Want a few things

- Parsing to get the groups so that `match.groups` is strongly typed
- Validating a regex?
- Maybe regex methods could be constant functions that can be evaluated at compile time?

Maybe use a external library? Needs to be ecmascript compliant

lemueldls:
Fixes #183
Fixes #175

image


checker/Cargo.toml
```diff
@@ -39,6 +39,7 @@ path-absolutize = { version = "3.0", features = ["use_unix_paths_on_wasm"] }
 either = "1.6"
 levenshtein = "1"
 ordered-float = "4.2"
+regress = { version = "0.10.0", features = [] }
```
kaleidawave, line 42: Looks like an awesome library!

checker/definitions/overrides.d.ts
```diff
@@ -201,6 +201,29 @@ declare class String {
...
 
+declare class RegExp {
+    @Constant("regexp:constructor")
+    constructor(pattern: string, flags?: string);
+
+    @Constant("regexp:exec")
+    exec(input: string): RegExpExecArray | null;
+}
+
+interface RegExpExecArray extends Array<string> {
```
kaleidawave, line 212: Man JS spec is crazy. So the returned result is an array with a bunch of added string/named properties! I think this definition where an interface extends a class works 🤞

checker/src/features/constant_functions.rs
```diff
@@ -306,6 +312,153 @@ pub(crate) fn call_constant_function(
...
+		}
+		"regexp:exec" => {
+			let this = this_argument.get_passed().map(|t| types.get_type_by_id(t));
+
+			if let Some(Type::SpecialObject(SpecialObjects::RegularExpression(regexp))) = this {
+				let pattern_type_id =
+					arguments.first().unwrap().non_spread_type().expect("pattern");
+				let pattern_type = types.get_type_by_id(pattern_type_id);
+				let pattern = {
+					let Type::Constant(Constant::String(pattern)) = pattern_type else {
```
kaleidawave, line 348: This is great, if the `exec` is run with a constant string argument then we get a constant result.

I assume [this example](https://www.typescriptlang.org/play/?ssl=7&ssc=2&pln=1&pc=1#code/JYMwBAFAdgpg7mASjA5gUQB4AcICIAWwuAlAHQwYwDGehJxYA3gFDMC+roksCy62tImQrU8AIwCGAE3pNWbIA) gets always true or false warnings under this branch (where TSC doesn't)

I wonder whether this can be extended for general (ie via Type::RootPolyType or Type::Constructor) strings not through the constant result but so that `.groups` has named members. So `new RegExp("Hi (?<name>.*)").exec(event.target.value)` would have groups as { name: string }. I think TS added this, but I can't find it?

checker/src/types/regexp.rs
```diff
@@ -0,0 +1,91 @@
...
+		} else {
+			format!("/{pattern}/")
+		};
+
+		let mut f = Flags::default();
+
+		if let Some(flags) = flags {
+			for flag in flags.chars() {
+				match flag {
+					'd' => unimplemented!("indices for substring matches are not supported"),
```
kaleidawave, line 28: Is this not something `regress` supports?

checker/src/types/regexp.rs
```diff
@@ -0,0 +1,91 @@
...
+					'g' => unimplemented!("stateful regex is not supported"),
```
kaleidawave, line 29: I think at the moment this can be ignored rather than panicking.

I think the *eventful* design of Ezno can support this in the future. For example evaluating a stateful RegExp could set a `regexp.#used = true` property. That would be a diagnostic that I don't think any other tool does today!

checker/src/features/constant_functions.rs
kaleidawave: Oh wait, I think this is being collected below? https://github.com/kaleidawave/ezno/pull/184/files#diff-be53d15efbd9ffdb15e107f75ef4867a569eb1ba61ecbd6cc1c468d3f49e8193R53

I will have a think about how this could work

checker/src/synthesis/expressions.rs
```diff
@@ -88,8 +88,8 @@ pub(super) fn synthesise_expression<T: crate::ReadFromFS>(
 		Expression::StringLiteral(value, ..) => {
 			return checking_data.types.new_constant_type(Constant::String(value.clone()))
 		}
-		Expression::RegexLiteral { pattern, flags: _, position: _ } => {
-			return checking_data.types.new_regex(pattern.clone());
+		Expression::RegexLiteral { pattern, flags, position } => {
+			return checking_data.types.new_regex(pattern, flags, position);
```
kaleidawave, line 92: 👍 nice. RegExp constructor and regex literal have the same logic

kaleidawave:
Nice! Didn't expect this so soon after I wrote that issue. This is great, really impressed that you got it a nice working example. It is a great use case for `constant_functions`, where the logic extends something representable in JS.

Minor things that I haven't already brought up in the review
- Could `types/regexp` be renamed/moved to `features/regexp`?
- Is is it possible to test whether the syntax is valid. Maybe `RegExp::new` could return `Result` (replacing the `.unwrap` here https://github.com/kaleidawave/ezno/pull/184/files#diff-be53d15efbd9ffdb15e107f75ef4867a569eb1ba61ecbd6cc1c468d3f49e8193R41)
- Somewhere near "constant operations" in specification.md, could you add a simple match example for RegExp (+any more independent features you think could be tested)

Looks great at the moment, if any of those things are too difficult LMK. Happy to just add the constant stuff as that is a really neat feature!

checker/src/types/regexp.rs
lemueldls: `regress` does not yet support all flags:
https://github.com/ridiculousfish/regress/blob/9f39399b0e61bee0602c791df8be468e00738100/src/api.rs#L59C1-L89C6

checker/src/types/regexp.rs
```diff
@@ -0,0 +1,91 @@
+use std::{collections::HashMap, fmt};
+
+use regress::{backends, Flags, Match, Regex};
```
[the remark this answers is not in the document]
lemueldls: It's hidden from docs, but I need it to access `CompiledRegex`:
https://github.com/ridiculousfish/regress/blob/9f39399b0e61bee0602c791df8be468e00738100/src/api.rs#L441C1-L444C19

lemueldls:
I've written a constant and variable version of the `exec` function, though I am not sure how to test non-constant strings (like `Type::RootPolyType` or `Type::Constructor`)..

I am experimenting with utf16 matching but am running into some correctness issues. `regress` is very standard compliant, so I'll find a test suite to find how to correctly match ranges.

I am also running into issues with optional property access of `RegExpExecArray | null` types on the result of the non-constant version of `exec`. Not sure if I am doing unions correctly.

kaleidawave:
> I've written a constant and variable version of the `exec` function, though I am not sure how to test non-constant strings (like `Type::RootPolyType` or `Type::Constructor`)..

Awesome. If you try
```ts
function greetingWho() {
	const r = new RegExp("Hi (?<name>.*)")
	print_type(r.exec("Hi Ben")!.groups)
}
```

Does it print `{ name : string }`? For the dynamic version is it creating a result `result = types.new_or_type(*result of object builder that contains special .group property*, TypeId::NULL)`

> I am experimenting with utf16 matching but am running into some correctness issues. `regress` is very standard compliant, so I'll find a test suite to find how to correctly match ranges.

Awesome, I think basic support is good for now. Maybe put a comment in `checker/README.md` that regex support is based on `regress` and mention any slight differences?

> I am also running into issues with optional property access of `RegExpExecArray | null` types on the result of the non-constant version of `exec`. Not sure if I am doing unions correctly.

From last week I think property access on an or `(left, right)` didn't check whether both left and right contained that property (aka `declare let a: { b: 2 } | null; print_type(a.b)` printed 2 rather than an error). I have just merged #157 (which is a huge PR where I redid a lot of the property things) which should have fixed that case (and also added non null assertions so the above should work once main is merged in).

kaleidawave:
Hey, putting out a release this week. Would love to have this feature in! Do you think you can maybe cut it back to what was working in the screenshot?

checker/src/features/constant_functions.rs
lemueldls: The TS Playground example works:

image

lemueldls:
Non-constant examples work, but I'm still not sure how I should write specification tests for it.

> Does it print `{ name : string }`?
It prints `{ name: "Ben" }`
image


image

kaleidawave:
Awesome!

> It prints { name: "Ben" } (but doing !. causes problems)

Yep its a case of #165. The parser thinks treats `!` as something before an expression and so it thinks the `.` means the start of a decimal number like `.25`. Will fix eventually. I think it works if you split it between lines with a variable.

> Only when I also remove the new_or_type does it give the expected result:

Hmm yes I think there are some problems with `?.` and narrowing at the moment...

kaleidawave:
Merged! 

Made a few changes before merge
- Fixed things `clippy` was whining about
- Added a function error for `new RegExp` with a bad input (should rename `ConstantFunctionError::BadCall` to `ConstantFunctionError::CannotComputeConstant` in the future)
- Added a test for bad expressions and the new bad function call above
- The ``Constant `RegExp.exec` groups greedy`` test was testing against a type checker error that printed the `.groups` object. However it looks like the order of items in `.groups` is unpredictable because regress uses a HashMap along the way. I have made the changes this side to use my own `crate::Map` for groups which I already use in the checker for small things and importantly has a stable order. To get stable ordering it needs a (hopefully small) change on the regress side: https://github.com/ridiculousfish/regress/issues/96. In the meantime I have changed the body of the test to not rely on ordering.

---
I did try adding `String.prototype.match` and `String.prototype.split` but there is something wrong with events on my side. I will investigate and let you when I get it working! Also I was worried it would add a big perf drop integrating this external process in, but looks okay so far!

PR #186, opened 2024-08-08
Basic operator checking

### LINKED ISSUE #92: Binary operator checking
Current the functions to doing binary operations don't check the types of either side. First a bit of background:

There are two functions for binary operators, they are split between
- Mathematical and bitwise operations. Both sides have to be checked as there is no short circuiting behavior
- Logical operators. These can have logical operators (and in the future the RHS can be narrowed for && operations)

*`in` and `instanceof` operators are handled specially on `Environment`*

There is a also a current option `strict_casts` (it should probably be renamed to `no_implicit_casts`) in `TypeCheckOptions`. When this is `true` should not allow operations like `"hi" + 2`. 

These two functions are currently set up to return a `Result`. However both the logic for checking the sides **and** the diagnostic has not been implemented. For example

https://github.com/kaleidawave/ezno/blob/502126217170121cadb12cad3debe1ff388a37f1/checker/src/behavior/operations.rs#L76-L84

There are several things to carefully consider here
- Still want to enable `Symbol.toPrimative` behaviour with hints and such (that will require these functions to have all the context to be able to do a `call_type`
    - This should **disallow** `Object.toPrimative` fall through as this result are not useful

kaleidawave:
- Basic operator checking (fixes #92)
- Inequality checks with ranges
- Identity checks from disjoint with `NaN`
- Disjoint with `Not` works (using subtyping)
- Proper checking that allows string + number?
- `infer T` in type annotation slice gets treated as `infer T extends string` (and also for generic arguments)
- Add `Number.isNaN` checks
- Change `GreaterThan` and `LessThan` to `InclusiveRange` and `ExclusiveRange`
- Add operations to ranges with new `FloatRange` type
- Changes `~`, `!` and `-` prefix operators to be in terms of `x ^ 0xFFFF_FFFF`, `x ? false : true` and `0 - x`. Removes `Constructor::UnaryOperation`
- Fixes to narrowing for `if (a)` where a is `string | falsy etc` and also to optional chaining
- Moves `a + b` is string or number to `result` field on `Constructor::BinaryOperator { result }`


kaleidawave (commented):
Pretty much done what I want for the initial functionality. Some things I might continue to add
- More checking for numbers
    - Operations on number intrinsics (might need to add result type for `Constructor::BinaryOperator`)
        - Has some problems caused by floats
    - Operations with zero, one, infinity, negative infinity
- More disjoint
    - Not intrinsic needs subtyping loop
    - Number intrinsics (also more in subtyping)
- `Math.sin`, `Math.random` types
    - These are inclusive inequalities though, not sure about `0 | 1 | (GreaterThan<0> & LessThan<1>)`  :/. Maybe a new intrinsic for a inclusive range? Both seem bloat
- Fixes and more
    - `Number.isNaN` narrowing cases
    - `Array.isArray` doesn't work because of Array as free variable (used before class declared) (maybe do before hand)

kaleidawave:
Will have a think about later if there is still stuff I want to add/change before merging tomorrow. But pretty happy with it for now.

kaleidawave:
Might have to change the range of `Math.random`. Turns out it is `[0, 1)` not `[0, 1]` 🤦‍♂️. Could do `0 | ExclusiveRange<0, 1>`? Will have to check that works

image

kaleidawave:
Merging before I get carried away with range division etc. Checking tests now up to 344 🎉

PR #187, opened 2024-08-08
Add iteration and related

kaleidawave:
- Add basic destructuring into array declaration based on iterator protocol object

TODO
- Destructuring assignment
- `Symbols` -> `Symbol`
- Handling errors
- `for of`

Maybe
- Generators?
- Promises?
- Length calculation
- Spreads
- `Array.prototype[Symbol.iterator]`

### DIFF
--- a/checker/specification/staging.md
+++ b/checker/specification/staging.md
@@ -1,3 +1,28 @@
 Currently implementing:
 
 > This file is for work-in-progress and can help separating features that are being implemented to regressions
+
+### Iteration
+
+#### Destructuring with iterating
+
+```ts
+const [x, y, z] = {
+    [Symbols.iterator]() {
+        let i = 0;
+        return {
+            next() {
+                if (i++ === 4) return { done: true, value: i };
+                return { done: false, value: i };
+            },
+            return() {
+                return { done: true };
+            },
+        };
+    }
+};
+
+({ x, y, z }) satisfies string;
+```
+
+- Expected string, found { x: 1, y: 2, z: 3 }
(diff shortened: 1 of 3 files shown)


checker/src/types/calling.rs
```diff
@@ -820,10 +820,17 @@ fn call_logical<B: CallCheckingBehavior>(
 						top_environment,
 						types,
 					);
-					let reflects_dependency = if types.get_type_by_id(result_as_type).is_constant()
-					{
-						None
-					} else {
+
+					let should_create_dependent_output = !types
```
kaleidawave, line 824: @lemueldls found the fix for `function newRegExp(name: string) { return new RegExp(name) }`. Turns out for `new RegExp` with some non-constant string it creates a new object with regexp prototype (bc it is a constructor), which I think is fine. But the problem is because it is just an object, it is considered constant and so didn't generate a non-constant result for the function call. Now fixed by changing adding this condition to account for the edge case.

kaleidawave: Would be cool to get `String.prototype.match` and `String.prototype.matchAll` etc working but need to do some investigation into all the oddities of the `Symbol.match` stuff

Also just sent you an invite to the contributor and sponsor repository that contains the LSP if you want to try it out ([like I did earlier](https://x.com/kaleidawave/status/1828440978992025937). Repository here https://github.com/eznolab/super-ezno-and-lsp

kaleidawave:
Would love to merge some of the improvements now (for the release) but unfortunately I think a lot more could be done on this. I have tested this a bit on some projects and unfortunately while there are improvements it doesn't quite completely clear some event / control flow issues.

There are a few issues with the `parent` field handling on `SubstitutionArguments` but unfortunately the changes caused a bunch of regressions.

PR #191, opened 2024-08-16
Change lexing

kaleidawave:
This will be a long WIP PR to change how lexing works and experiment with some different ideas.

Currently the default mode from the lexer is a lexer in one thread that pushes tokens to a `mspc::channel` and another thread that reads

A few reasons
- There are limitations to `expect_expression` https://github.com/kaleidawave/ezno/issues/165
- The parser is currently slower than the later type checker phase. Could changes improve performance?
- Simplify things with new lines, partial syntax and lookahead. Allow WASM to be under the same mode and to not buffer the whole source (currently buffers everything rather than 

Things to change
- Move away in the lexer *emitting tokens*. Instead a lexer that the parser *requests tokens* from
- Allow the parser to request a different lexer mode based on the state of the parser
- Avoid intermediate tokens like `) ] }` and other things via requested from `expect_next`. 
- Avoid the `peek()` buffering system

Others (that might be other prs)
- Not using owned strings, instead references into a store (that may or may not be the FileSystem map from source-map)
- Using Vec<u8>/&[u8] instead of String/&str, that might reduce overhead in utf8 checker (or whatever)
- Changing token and AST definitions to be more performant (?)

Opening this as a PR at this early stage to track updates and not lose any progress.

> Could use something like [grit](https://docs.grit.io/) to automate large API changes?

> Again there are better parsers out there now. Not all feature complete and as small. This isn't something that is needed but interesting to experiment with and helps with patterns in the checker


Aetherall:
Hi ! I am surely telling nonsense, but maybe using an existing parser such as https://github.com/tree-sitter/tree-sitter/tree/master would allow you to reduce the scope of ezno ? There is a rust binding for tree-sitter if i am not mistaken. But maybe it does not suit your needs I really dont know.

Anyway much love for your work, I am really hyped to see the project really going forward

kaleidawave:
> Hi ! I am surely telling nonsense, but maybe using an existing parser such as [tree-sitter/tree-sitter@`master`](https://github.com/tree-sitter/tree-sitter/tree/master?rgh-link-date=2024-08-19T12%3A18%3A27Z) would allow you to reduce the scope of ezno ? There is a rust binding for tree-sitter if i am not mistaken. But maybe it does not suit your needs I really dont know.

tree-sitter is a great library! Haven't used in directly before I have used some products that use it. It does have the features that would be useful for the type checker: you can use it from Rust, it can produce a partial AST with missing syntax and I think you can turn it back into a string as a build step.

When I started the parser in early 2021 I didn't know much about alternatives and since a few more TypeScript/JavaScript parsers written in Rust have popped up. 

At the moment I think the best idea is to continue with the current code in `/parser`. It is pretty good and fast already. It supports partial syntax and some other experiments. While the type checker is designed to not contain any reference to specific AST, having it use `ezno-parser` by default is easier to work with as it is under the same repository and I am more familiar with the structure and APIs. 

This PR is more on the backburner (might be a few months away, not focusing on it RN), to experiment with some ideas I have to make the current parser simpler, more complete and faster. Not really because it is necessary but because I think it would be interesting to explore and it would be nice to the parser in a better shape.

I hoping this should require a small few tweaks and simple transformations. Learning a different AST library and rewriting the code under `checker/synthesis` is probably more work that this.

Thanks for the suggestion anyway!

> Anyway much love for your work, I am hyped to see the project really going forward

Thanks 🎉

kaleidawave:
Making some more progress on this:
- More compatibility, got a few more complex examples including the return type annotations in arrow functions inside a ternary conditional expression (https://x.com/kaleidawave/status/1837833122437632370)
- [Parsing is roughly 2x faster just on this initial change](https://github.com/kaleidawave/ezno/actions/runs/11031980468) (I think more could be done to improve this)

To add back
- Partial syntax
- Better errors (some are todos)
- The constant compilation stuff

To fix
- Some issues with type annotations being default
- Renaming some methods
- Some AST node positions are broken
- Fuzzing (I think because there are `panic!`s in place of actual errors)
- Tests (again maybe positions)

kaleidawave (commented):
Things are going well. Few things
- `partial_syntax` causes cycle and then runs out of memory. Example `h\n)-`
- Problems with char indexing. Example `<>q//º\u{c}`
- New non-null assignment needs to be under feature

Maybe should add more files to the CI parsing test

Will fix at some point

kaleidawave:
Is working pretty well. Can touch up any minutiae later

Issue #22, opened 2023-06-08
Add a LSP (language service protocol) extension


kaleidawave:
*need to update the 2k lines of code I wrote last year to the new Ezno*

[There is an example video near the top of the page](https://kaleidawave.github.io/posts/introducing-ezno)

Things to consider:
- AST has to be partially valid, this may only work for `ezno-parser` as it has [Cursor](https://docs.rs/ezno-parser/0.0.3/ezno_parser/expressions/enum.Expression.html#variant.Cursor) variants which enable `const x = ;` to be valid AST 
- Distribution using WASM seems like a good solution: https://www.osohq.com/post/building-vs-code-extension-with-rust-wasm-typescript
- What features of https://microsoft.github.io/language-server-protocol/ to support

Future things
- Not rechecking the whole source every write

kaleidawave:
This probably won't happen this year as I don't think it is particularly useful while the checker is still in demo stages.

Maybe I will release the LSP I build last year but it will likely only be for sponsors to test and only print diagnostics without completions etc...

Or this feature gets added if Oxc gets a vscode extension? 👀  https://github.com/web-infra-dev/oxc/issues/688

Boshen:
We have the code here https://github.com/web-infra-dev/oxc/tree/main/editor/vscode

Need to find the time to publish this thing :-/

kaleidawave:
A basic LSP is now available to GitHub sponsors. It is similar to the [web playground](https://kaleidawave.github.io/ezno/playground/) but works locally (in vscode) and can support multiple files & imports etc. Similar to the checker it is WIP & not production ready. It will be available only to sponsors for the foreseeable future.

If you have contributed code to this repository and want access then LMK and I can add you to the private repository!

PR #202, opened 2024-08-30
Add watch mode support

### LINKED ISSUE #164: Watch mode
It would be great to have a `--watch` mode for the `ezno check` command. Effectively running `ezno check` anytime a file in the project is updated.

notify: https://docs.rs/notify/latest/notify looks like a good library to do this!

Currently the checker doesn't do anything incremental. So it is fine just to call `check_project` again and again without any fine grained specifics for only part of a project to be checked / invalidated.

One consideration is whether then watch will support glob syntax (that is coming in #155)? Maybe the watched file could be run through [Pattern::matches ](https://docs.rs/glob/latest/glob/struct.Pattern.html#method.matches)

Tommypop2:
Closes #164 

This PR is much bigger than I intended it to be, and clones data each time a file change is observed. This could possibly be removed by changing the underlying API, but I'm honestly not sure.

Also, on every individual file change, every file that matches the glob is recompiled. I did this for 2 reasons:
- The compiler might take multiple files into account (didn't look into the actual compiler source code so wasn't sure if this was the case, and if it doesn't now, it might in the future)
- If only the changed file is rerun, then the diagnostics for the rest of the project wouldn't be reprinted. This could maybe be solved by caching the last output for each file, and just reprinting that if the file hasn't changed

My current way of supporting the glob syntax is just by using the `entry_points` provided by `get_entry_points`. This seems to work pretty well, and reuses the work that needed to be done anyway to get the list of entry points. However, I'm not sure how efficient individually watching all the files is over many files.

If anything needs changing, let me know! Or if this approach is terrible I'm happy to close this


checker/src/options.rs
Tommypop2, line 1: I didn't particularly want to do this, since it required changing the internal checker code, but it seemed like the only way to get it to work as ownership is passed to the checker when it's passed in.

kaleidawave:
Awesome, am back next week will have a closer look then.

Tommypop2:
Ok, thank you!

kaleidawave:
Yes at the moment the checker doesn't support anything incrementally. In fact the web playground https://kaleidawave.github.io/ezno/playground and the LSP #22 run `check_project` on every keystroke (-some throttling). It can get away with it for the moment because the check process is fast. I want to work on some partial file and possible even local in a file rechecks, but it is quite low priority. So at the moment a watch mode that runs the `check_project` is good enough.

I will double check this and it will definitely be in the next release but that may be a while as currently a bit busy with things. Thanks for the addition though :)

kaleidawave:
Have got it compiling under WASM now. Will check on it later. 

I think a sort of "initial" version would be good to include in the next release. I can't think of a good way to add a test for it ATM

Issue #198, opened 2024-08-27
Add `Object.seal` and `Object.preventExtensions`


kaleidawave:
Currently `Object.freeze` works via `frozen` on `LocalInformation`. This current system could be extended to  account for these additional two object modifications. The frozen on `LocalInformation` should be changed from a `HashSet` to `HashMap<TypeId, Frozen | Sealed | NoExtensions>` and various lookups should be done on this for certain usages of objects.

See:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/seal
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/preventExtensions

PR #213, opened 2024-10-30
feat: support `Object.seal` object `Object.preventExtensions`

sor4chi:
Hello! Thank you for this fantastic open-source project!

I was intrigued by Ezno, so I decided to try one of the good-first-issues to study and get familiar with the codebase.

---

#198

- update checker type definitions (and generate binary)
- define the frozen type for multiple freezing ways
- extend the frozen field of LocalInformation with frozen type

As far as I know, objects that have been `Object.freeze`, `Object.seal`, or `Object.preventExtensions` will all have both `Object.isFrozen` and `Object.isSealed` return true. Therefore, I’m using existence checks based solely on frozen HashMap keys.


kaleidawave:
Awesome! Thanks for checking it out and even more for the contribution. Currently a bit busy this afternoon (you will see soon) so will check this out tomorrow or Friday. You can ignore the CI errors for now as they are for separate things that I will fix.

In the meantime you can check out the specification.md file to see how the type checker is "unit tested". If you have any of the new features working right now for `seal` and `preventExtensions` and it catches errors not currently presently caught in the main branch it would be good to add some custom tests in the markdown format of specification.md (but put them in the staging.md file for now to keep them separate from the finished features).

Thanks 🙏

kaleidawave:
Awesome! This looks ready to merge. I will have to re-read the MDN documentation on this to double check that everything is correct and whether there is anything else that could be implemented. But looking good, new specification tests passing AND no breakages 🎉  https://github.com/kaleidawave/ezno/actions/runs/11603282627/job/32383499515?pr=213#step:8:323

sor4chi:
I read the MDN and noticed that `Object.isExtensible`. (false if any of the three fixing methods are used).

I'll implement this as well.

https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/Object/isExtensible

sor4chi:
How about this? 
Now it seems unlikely that there will be any more to add.

kaleidawave:
Awesome! Will merge later.

kaleidawave:
Hey! Great work, I like the abstraction with `enum ObjectProtectionState` and nice work figuring out `constant_functions`. Made a few changes
1. Fixed my own mistake as was a single line: in assignment, it was looking at the protection state in the current environment but not ones above (using `facts_chain`), it now uses the proper method so it should not allow to modifying a frozen object in an above context. For example `const x = Object.freeze({}); if (condition) { x.g = 2; }` now throws a type error.
2. The seal allows modifying existing properties, so I changed the logic to only throw a type error if registering a new property (updated the tests to match that behaviour)
3. Also changed the constant functions to be a little more specification correct

> There is still some uncaught edge cases with `Object.isFrozen(Object.seal({})) === true` because `.isFrozen` seems to observe existing properties and whether any are configurable. https://natto.dev/@kaleidawave/12d97430628e4e8ebe5408b1e67992a8. Don't fully understand ATM, but should be fine for what I really wanted which is `.seal` and `.preventExtensions`

Also did what I do for every PR when they are finished which is to merge the staging tests into the final specification once it is ready! (I should come up with a way to automate that process).

kaleidawave:
Also thanks for the sponsorship 🎉. Next weeks blog post should be about properties, so I will mention this feature in it!

sor4chi:
Thank you very much!
The code was very easy to understand and debugging was a breeze.
I will start on the next task (good-second-issue) before this interest dies down!

Issue #196, opened 2024-08-23
Generic names leak out of functions


kaleidawave:
Somehow type parameter `YEA` has its name declared in the main scope (not the function scope)

```ts	
function x<YEA>() {}

type B = YEA;
```

This might have to do with the hoisting (`synthesise_shape`)

sor4chi:
Hello, I investigated this issue!

It seems that in the `hoist_statements` function, specifically in the Function match section, `synthesise_shape` is applied to the environment outside the function scope without generating a new lexical scope within the function. As a result, it might be registering to the `named_types` in the scope one level higher than intended.


https://github.com/kaleidawave/ezno/blob/d67cd1b0b9f5c308a8d3f8ea04c5ccb8baa470c5/checker/src/synthesis/hoisting.rs#L559-L568

What do you think?

kaleidawave:
Yes, I think this is the issue. Btw `synthesis_shape` is hoisting the function type with its value parameter types and type parameters (type parameter names are the ones leaking).

The stages for *checking* (filtered for functions) 
1. `synthesis_shape` type annotations on a function
2. `synthesis_function` body of a function
3. *run statements*

I think you are correct, the same environment is used for `synthesis_type_parameter` 

https://github.com/kaleidawave/ezno/blob/d67cd1b0b9f5c308a8d3f8ea04c5ccb8baa470c5/checker/src/synthesis/functions.rs#L636-L641

I think it needs to be modified to do a similar process to function type annotations where it creates a new lexical environment and does the synthesis inside.

https://github.com/kaleidawave/ezno/blob/d67cd1b0b9f5c308a8d3f8ea04c5ccb8baa470c5/checker/src/synthesis/functions.rs#L597-L602

Would you be interesting in committing and adding a test for the fix?

sor4chi:
Thank you! Of course, I'll try.

PR #216, opened 2024-11-06
fix: synthesize shape in sub-environment scope

sor4chi:
Hi, I resolved generic names leaking. I'm not sure if I'm right...

#196

---

- wrap shape-synthesizes operation with new inner lexical scope
- add an edge test case


kaleidawave:
Awesome! I think I will be around to merge this later

kaleidawave:
Sorry dealing with college stuff and don't want to rush a release. Will be in next week!

Issue #217, opened 2024-11-06
Number-embedded template literal doesn't validate collect


sor4chi:
## TSC

in:
```ts
const invalidNum: `${1}` = 1;
const invalidNum2: `${1}` = "1";
const invalidNum3: `${1}` = "2";
```

out:
```txt
tmp.ts:1:7 - error TS2322: Type '1' is not assignable to type '"1"'.

1 const invalidNum: `${1}` = 1;
        ~~~~~~~~~~

tmp.ts:3:7 - error TS2322: Type '"2"' is not assignable to type '"1"'.

3 const invalidNum3: `${1}` = "2";
        ~~~~~~~~~~~
```

## Ezno

in:
```ts
error: 
  ┌─ tmp.ts:2:29
  │
2 │ const invalidNum2: `${1}` = "1";
  │                    -----    ^^^ Type "1" is not assignable to type string
  │                    │         
  │                    Variable declared with type string

error: 
  ┌─ tmp.ts:3:29
  │
3 │ const invalidNum3: `${1}` = "2";
  │                    -----    ^^^ Type "2" is not assignable to type string
  │                    │         
  │                    Variable declared with type string
```

kaleidawave:
Fixed in #209

PR #218, opened 2024-11-09
feat: enhance template literal type printing

### LINKED ISSUE #188: Template literal type printing
At the moment the template literal **type** `` `Hi${string}` `` is printed at `string` (as that is its base type).

There should be special logic to pick up the type `Constructor::BinaryOperation { lhs, rhs, operation: MathematicalAndBitwise::Addition }` and then form a chain of `parts: Vec<(String, TypeId)>, trailing: String` using recursive type traversal. If it looks like a template literal type (aka no number + number) then it can be printed like on (see example in parser)

It should be added around here

https://github.com/kaleidawave/ezno/blob/e6f6ffd7cb48285cdaaf7e57e4d636bbaf038b03/checker/src/types/printing.rs#L254-L281

sor4chi:
Print the constituents of template literal types. Fixes #188

### DIFF
--- a/checker/specification/specification.md
+++ b/checker/specification/specification.md
@@ -2441,7 +2441,7 @@ fakeRead(array1)
 #### Always known math
 
 ```ts
-// True regardless of 
+// True regardless of
 function func(a: number) { return a ** 0 }
 
 func satisfies string;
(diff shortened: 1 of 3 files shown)


kaleidawave:
Awesome! Now I need to fix #217 !

PR #226, opened 2025-01-31
Fix RegExp UTF-16 Matching

lemueldls:
The crate used to evaluate regexp, [regress](https://github.com/ridiculousfish/regress), fixed the [utf-16 matching problems](https://github.com/kaleidawave/ezno/issues/184#issuecomment-2274383051) in their [0.10.3 release](https://github.com/ridiculousfish/regress/releases/tag/v0.10.3).

The `RegExp::exec` function now correctly matches using utf-16 instead of utf-8.

### Before (`main`):
snapshot_2025-01-31_10-36-01

### After (`regexp-utf16`):
snapshot_2025-01-31_10-38-35

### Expected:
snapshot_2025-01-31_10-45-05

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,18 +3,6 @@
 version = 3
 
 [[package]]
-name = "ahash"
-version = "0.8.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


kaleidawave:
Awesome! Will look at merging next week.

PR #230, opened 2025-05-26
Start on new version of specification

kaleidawave:
This is the start of a new specification

Things I would like to improve on the existing specification
- Include more of the whole project (parser, etc)
- Split the current 5k file into smaller parts
- Add more information around certain topics
- A few more tests here and there for certain behaviour
- Also act as documentation for the project
- Published on the web alongside the playground

Currently running the code blocks is done by a `build.rs` and some helper methods. I think it would be cool to build out a general testing tool based on markdown. However, that is something more for the future. Initially we can do something similar to before but not using `build.rs` and generating a bunch of Rust code.

Some first things to complete
- [ ] Merge existing "documentation"
- [x] Merge existing tests
- [x] Enumerate file names
- [ ] Think about layout more
- [ ] Move to respective crate directories and use symbolic links

Then
- [ ] More tests
- [ ] Lint and spell check markdown files
- [ ] Publishing to HTML


kaleidawave:
Next up
1. Enumerate files giving more order and structure
2. Get testing working
3. Get amalgamate script working with performance
4. Add workflow to check markdown formatting
5. Merge existing *documentation*
6. Merge existing `to_implement.md` with some sort of marker
7. Remove existing `specification.md` (once we know it is 100% working)
8. Replace `staging.md` behaviour
    - Define some `unsorted.md` or `work-in-progress.md`
    - Use `git diff`?
9. More files and more information inside of them
10. ...

PR #232, opened 2025-07-03
Expose type checker API

kaleidawave:
Adds a basic API for working with types

```rust
let mut space = Space::new();

assert!(space.is_assignable_to("number", "number | string"));
assert!(!space.is_assignable_to("number | string", "number"));
```

#### TODO

- Candidates from inference
- More mismatch information
- Disjoint tesy
- Declare types and variables
- WASM API


kaleidawave:
Made some progress but waiting on the offset fix in #238

PR #238, opened 2025-10-29
Fixes to parser

kaleidawave:
More simplifications and test262 improvements

- switch case `Expression` -> `MultipleExpression`
- Improve skipping
- Yield as function
- Decorators
	- `var x = @decorator`
	- Fixes to yield and await names 
- Comments
	- Skip comments everywhere
	- Remove `WithComment`
	- #TODO what to do about comment thing
- Fix issues with unicode surrogates
- identifiers use match indicies, proper unicode_ident continue property and string unicode escape function
- `export ... from ... with { ... }`
- `and`, `or` and `if` for nunjucks
- `try_parse`
	- For arrow functions
- `expect` -> `expect_chr`
- replace if-chains to match (using east)
- Update `trait ASTNode`
	- Allow partial source
	- `offset: Option<usize>` -> `offset: usize`
- Update checker for above things
- Backtracking for `FunctionHeader` and `MethodHeader`


kaleidawave:
Want to have a look at where these "parsed when should have failed" checks can be done. Also want to remove lookahead in for loops. Just parse the start then figure out afterwards if not at `=`.

A few other issues we should have a look at as well

kaleidawave:
Still to do

- More testing (parser [spectra](https://github.com/kaleidawave/spectra) tests, may need the [options PR](https://github.com/kaleidawave/spectra/pull/3) branch)
- More lookahead. Some items imperfect performance wise, but we can come back at a later date
- Changing branching in `Expression::first_item`, `TypeAnnotation`

PR #241, opened 2026-06-16
Report TDZ writes to free variables

### LINKED ISSUE #239: Add TDZ check for event writing to a variable
https://github.com/kaleidawave/ezno/blob/8a763a0a1e92317b4822a9ea1cfaaf150b036f12/checker/src/events/application.rs#L145

This line should raise a reference error if the yielded value `is_none()`  with the same logic as used for read variable 

https://github.com/kaleidawave/ezno/blob/8a763a0a1e92317b4822a9ea1cfaaf150b036f12/checker/src/events/application.rs#L97-L107

A valid PR would include a single new test for this next to the read variable one 

https://github.com/kaleidawave/ezno/blob/8a763a0a1e92317b4822a9ea1cfaaf150b036f12/checker/specification/specification.md#tdz-from-free-variable-across-function

WilliamK112:
## Summary
- Report `VariableUsedInTDZ` when applying a `SetsVariable` event to a free variable whose current value is unavailable.
- Keep applying the write after recording the diagnostic, matching the existing read-side behavior that avoids cascading failures.
- Add the requested specification example next to the existing free-variable TDZ read case.

Fixes #239

## Validation
- `rustfmt +stable-x86_64-pc-windows-gnu --edition 2024 --check checker/src/events/application.rs`
- `cargo +stable-x86_64-pc-windows-gnu check --workspace`
- `cargo +stable-x86_64-pc-windows-gnu check --bin ezno`
- `cargo +stable-x86_64-pc-windows-gnu test -p ezno-checker`
- `cargo +stable-x86_64-pc-windows-gnu run -p ezno-checker --example run_checker ../tdz-write.ts --simple-diagnostics`
- `cargo +stable-x86_64-pc-windows-gnu run -p ezno-checker --example runner -- --interactive --rpc` for both TDZ-write and declared-before-write cases

Notes:
- I couldn't run the exact full `spectra test specification/specification.md ...` command locally because the `spectra` release asset referenced by CI was not available in `kaleidawave/experiments@assets`; the runner protocol used by spectra was validated directly for this case.
- `cargo clippy -p ezno-checker -- -A warnings` currently fails under Rust 1.96 on existing lints outside this change, such as `parser/src/lexer.rs` and a few checker files.

### DIFF
--- a/checker/specification/specification.md
+++ b/checker/specification/specification.md
@@ -1614,6 +1614,20 @@ let x: number = 5;
 
 > Not shown in the example but thanks to [#69](https://github.com/kaleidawave/ezno/pull/69) for adding the position of the error
 
+#### TDZ from free variable write (across function)
+
+```ts
+function setX() {
+	x = 1
+}
+
+setX();
+
+let x: number = 5;
+```
+
+- Variable 'x' used before declaration
+
 #### TDZ errors through nested getter
 
 ```ts
--- a/checker/src/events/application.rs
+++ b/checker/src/events/application.rs
@@ -123,33 +123,34 @@ pub(crate) fn apply_events(
 					// TODO maybe needs to be returned back up, rather than set here?
 					top_environment.possibly_mutated_variables.insert(*variable);
 				} else {
-					if get_value_of_variable(
-						top_environment,
-						*variable,
-						Some(type_arguments),
-						types,
-					)
-					.is_none()
-					{
-						diagnostics.errors.push(
-							crate::types::calling::FunctionCallingError::VariableUsedInTDZ {
-								error: VariableUsedInTDZ {
-									variable_name: top_environment
-										.get_variable_name(*variable)
-										.to_owned(),
-									position: *position,
-								},
-								call_site: input.call_site,
-							},
-						);
-					}
-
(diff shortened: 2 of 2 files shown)


kaleidawave:
Thanks for the quick change!

I have tested it locally. Apologies the testing is a bit old and broken. Will update in #240 

I got [spectra](https://github.com/kaleidawave/spectra) (for macOS) from `gh release download -R kaleidawave/spectra v0.1.0 --pattern '**aarch64**' -O spectra`. I added the frontmatter `lists-as-code-blocks: true` and then ran testing.

Aside from some of my own issues around modules being broken (for some reason), you additional test passed!

Unfortunately, there were some new false positives. 

Moving the check to only happen if `in_above_environment` is `true` removed those false positives though while keeping the new test passing!

I have put in a commit that fixes this

Otherwise: I think your use of `get_value_of_variable` is better because it does not suffer from a false positive case that my `let defined = info.variable_current_value.insert(...).is_none()` would encounter in the following

```typescript
let x: number = 0;
function u() { x = 2 } 

if (condition) {
  // the if branch context does not have a *current value* for *x*
  u()
}
```

Once I have fixed testing upstream and tested a few other cases (e.g. closures) I will merge. 👍

